PR #19, opened 2021-07-11
Config Dir Structure Proposal

NicolaBernini:
The `config/config.yaml` is going to be the master config file with different sections 
The 2 main sections proposed here are 
- `web3` containing the Web3 related information like the Infura Node used to connect 
- `inspection-target` containing the information about the target smart contracts meaning their address and a pointer to the ABI file


bertmiller:
Closing this as we have a config setup

PR #20, opened 2021-07-11
Dockerization_Proposal

NicolaBernini:
Dockerization Proposal 

Usage 

1. Build the Docker Image 

Run `build.sh`

2. Get a shell in the container 

Run `enter.sh`

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.6
+FROM python:3.9
 LABEL maintainer "Nicola Bernini <nicola.bernini@gmail.com>"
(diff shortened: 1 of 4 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM python:3.6
```
lukevs, line 1: Any reason not to go to latest stable release? (3.9?)

NicolaBernini: No, it is just my bias of being a bit conservetive with respect to new stuff 
Will update to 3.9

bertmiller:
LGTM

Issue #2, opened 2021-07-07
Create a file for data classes


bertmiller:
In particular we need to finalize and standard "actions," which are the output of inspectors. Then this can be used to better handle the output of all inspectors.

I imagine the standard practice is to throw all new data related classes in one file and import that where needed, but I could use some guidance here

bertmiller:
@lukevs is tackling this task

PR #21, opened 2021-07-12
Add pydantic Block

lukevs:
## This PR
A first step towards https://github.com/flashbots/mev-inspect-py/issues/2

- adds a pydantic implementation for BlockData (as Block)
- reimplements block.py to use it
- adds a structure for adding schemas for future data definitions
- fixes a typo
- adds a requirements.txt for dependencies

Also fixes a bug in how we were encoding HexBytes objects (they were becoming `{}` b/c of `.__dict__`)
image

The children of Block are left as dicts / list of dict right now
Leaving for future PRs to keep this first one focused

## Reading roadmap

- start with `schemas/blocks.py` to see how the object is defined
- then look through usage / changes in `block.py`

## Why pydantic v. dataclasses v. attrs etc
Pydantic:
- benchmarks as the fastest option right now: https://pydantic-docs.helpmanual.io/benchmarks/
- does runtime validation that all fields are set + are the right type
- has simple methods for JSON serialization / deserialization
- is generally nice to use!

## Testing

- [x] verify testing_file.py runs for a cached block and an uncached block
- [x] delete existing cache/*.json, run testing_file.py to bring each back, diff with the old (found the bug mentioned above)
- [x] verify that all cache blocks currently in the repo are unchanged before and after read / write via a Block
```
cache_directory_path = Path(cache_directory)

for filepath in cache_directory_path.iterdir():
    with open(filepath, "r") as file:
        json_dict = json.load(file)

    block = Block.parse_file(filepath)
    block_dict = json.loads(block.json())

    assert json_dict == block_dict

print("All good!")
```

### DIFF
--- a/block.py
+++ b/block.py
@@ -99,5 +99,5 @@ def createFromBlockNumber(block_number: int, base_provider) -> Block:
 
 
 def _get_cache_path(block_number: int) -> Path:
-    cache_filepath = f"{cache_directory}/{block_number}-new.json"
-    return Path(cache_filepath)
+    cache_directory_path = Path(cache_directory)
+    return cache_directory_path / f"{block_number}-new.json"
--- a/block.py
+++ b/block.py
@@ -2,7 +2,7 @@ from web3 import Web3
 from pathlib import Path
 import json
 
-cache_directoty = './cache'
+cache_directory = './cache'
 
 class BlockData:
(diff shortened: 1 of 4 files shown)


bertmiller (approved):
LGTM! Ran through your tests after pulling your branch to my local machine. Thanks for this.

PR #22, opened 2021-07-15
Add BlockCall class

lukevs:
# This PR

Adds a data class for the "calls" field in Block
Also adds some generic utilities for helping in translating from camel case to Python-standard snake case
Also moves the logic for building a new block to `fetch_block` to use for manual testing (+ cleaner)

Writing this:
- to make it easier for inspector implementors to see what data they have available in calls (still TODO actions + result types)
- as a first example of adding a child type on Block + handling the awkwardness around camel case

Todo in future PRs:
- add typing for the different actions and results for each call type
- adopt these in the inspectors - to make things easy for now I switched the call objects back to JSON when passing to inspectors so they don't need to update, but this should change to use the objects when we're ready

# Testing

Ensured there are no validation errors from running against different ranges of block numbers
```
    base_provider = Web3.HTTPProvider(rpc)
    w3 = Web3(base_provider)

    for i, block_number in enumerate(range(block_start, block_start + block_limit)):
        typer.echo(f"Testing {block_number} ({i+1}/{block_limit})")
        # includes trying to load the block
        fetch_block(w3, base_provider, block_number)
```


bertmiller:
Running the following:
`robertmiller@Roberts-MacBook-Pro mev-inspect-py % python3 testing_file.py -block_number=12768559 -rpc=RPC_ENDPOINT`

Gets the following output:
```Traceback (most recent call last):
  File "testing_file.py", line 1, in <module>
    from processor import Processor
  File "/Users/robertmiller/Documents/GitHub/mev-inspect-py/processor.py", line 1, in <module>
    from schemas.utils import to_original_json_dict
ModuleNotFoundError: No module named 'schemas.utils'```

I ran
`pip3 install schemas`

Then I got the following output with the same command as above
```Traceback (most recent call last):
  File "testing_file.py", line 1, in <module>
    from processor import Processor
  File "/Users/robertmiller/Documents/GitHub/mev-inspect-py/processor.py", line 1, in <module>
    from schemas.utils import to_original_json_dict
  File "/Users/robertmiller/Library/Python/3.8/lib/python/site-packages/schemas/__init__.py", line 79
    print "Schema violation for key '{0}' and value '{1}'".format(k, v)
          ^
SyntaxError: invalid syntax```

Am I doing something wrong here? I've got everything in requirements.txt

lukevs:
@bertmiller hmm - I’ll rerun it from the docker container today so we have consistent environments

schemas.utils comes from `schemas/utils.py` so that import should work regardless of packages

bertmiller:
I dug into it and my `schemas\utils.py` got deleted the first time! Not sure how or why. Restoring those it worked fine!

PR #23, opened 2021-07-15
Add project structure

lukevs:
Moves all mev-inspect source code to a module `mev_inspect`
Leaves scripts, docker, cache, and tests at the top level

Following this standard pattern: https://docs.python-guide.org/writing/structure/

Makes it clearer what is and is not part of the main source code
Also makes it easier for if / when we decide to make this repo pip-installable

## Testing
- [x] Verified testing_file.py still works
- [x] Verified all tests pass running `python -m unittest tests/tokenflow_test.py`


bertmiller (approved):
Pulled and ran locally on a block successfully, LGTM and changes make sense

PR #25, opened 2021-07-15
Add types for ABIs

lukevs:
Adds some typing around loading ABIs
Also adds convenience functions for creating function signatures and selectors
I only added the fields I'm using so far. More can be added as needed

## Testing
Script
```
from mev_inspect.abi import get_abi


def main():
    router_abi = get_abi("UniswapV2Router")
    add_liquidity_function = router_abi[2]
    print(add_liquidity_function.name)
    print(add_liquidity_function.get_signature())
    print(add_liquidity_function.get_selector().hex())


if __name__ == "__main__":
    main()
```
Output
```
addLiquidity
addLiquidity(address,address,uint256,uint256,uint256,uint256,address,uint256)
0xe8e33700
```

### DIFF
--- a/mev_inspect/abi.py
+++ b/mev_inspect/abi.py
@@ -1,0 +1,20 @@
+import json
+from typing import Optional
+
+from pydantic import parse_obj_as
+
+from mev_inspect.config import load_config
+from mev_inspect.schemas import ABI
+
+
+ABI_CONFIG_KEY = "ABI"
+
+config = load_config()
+
+
+def get_abi(abi_name: str) -> Optional[ABI]:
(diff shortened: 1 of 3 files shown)


bertmiller:
Helpful! LGTM

bertmiller:
Also, just as a note, you can also do this in web3.py

```
Greeter = w3.eth.contract(abi=abi, bytecode=bytecode)
Greeter.functions.getTime().selector
```

However, I found that the selector generated by this didn't match the type of the calls, and I initially struggled to make that work, so instead I opted to generate the selectors myself using Web3.sha3.

PR #26, opened 2021-07-20
Add NestedTrace

lukevs:
## This PR
Adds a NestedTrace object and the ability to build NestedTraces from a list of Traces

A NestedTrace represent the tree of traces expressed by the "trace addresses" of a trace and it's subtraces (say trace again)

Example:
A parent trace at address [0, 0, 1] might have subtraces at [0, 0, 1, 0], [0, 0, 1, 1, 0], etc because they share a prefix

We'll pass NestedTraces to inspectors to save them from building this tree themselves

## Also

- Renames BlockCall to Trace + calls => traces
- Deletes all cached blocks (they're not compatible with the traces name)
- `createFromBlockNumber` => `create_from_block_number`

## Testing
I wrote a unit test which is runnable with
```
python -m unittest tests/trace_test.py
```


bertmiller:
I needed to run testing_file.py once to cache block number `11931272`, but after that the unit tests worked great.

The problem is that tokenflow_test.py relies on a custom block loading implementation that assumes a cached block. We should have it use the standard `create_from_block_number` in `block.py`, which is an easy fix.

Could you quickly address that? Then feel free to merge!

PR #27, opened 2021-07-20
Add inspectors and classifications

lukevs:
## This PR

Adds an abstract base class defining the interface for an Inspector
Adds a base class for Classification
Reimplements the processor + uniswap inspector to satisfy this interface

## Testing
Verified that testing_file still runs
Also verified that the number of returned classifications == number of unique transactions


bertmiller:
Does this need to be merged after your other PR?

Looks good to me though!

PR #30, opened 2021-07-23
Add poetry

pmdaly:
Poetry solves a lot of common package management challenges and greatly simplifies development.  It's modeled after Rust's Cargo and very similar to Yarn.  The following changes were added in the migration to Poetry.

1.  Requirements are now in pyproject under dependencies and dev dependencies, versioning is now managed.  New dependencies can be added via `poetry add package` or `poetry add --dev package`
2. Python version file is now apart of pyproject, if the original structure is required by another package we can revert this, but I think this should be fine.
3. Scripts are now added in the dev_tools file and can be accessed via `poetry run lint' for example.  Commands include lint, test (uses pytest now), isort, isortcheck, mypy, black, blackcheck, and start, build and start_it which are docker commands
4. Pytest was added along with coverage and some "sugar" to "prettify" the output.  Pytest will work with unnittest so no changes are required but I would encourage tests to be written using pytest, it's far simpler.

Setting up a project is now as simple as `poetry install` which will install all the dependencies and setup an environment.  The env can be accessed via `poetry shell`.


Dockerfile
```diff
@@ -1,10 +1,20 @@
...
-CMD []
+RUN poetry install
+
+# easy bruh...
+#ENTRYPOINT ["./run.sh"]
+
+# easter eggs 😝
+RUN echo "PS1='🐳:\[\033[1;36m\]\h \[\033[1;34m\]\W\[\033[0;35m\]\[\033[1;36m\]$ \[\033[0m\]'" >> ~/.bashrc
+
+CMD /bin/bash
```
lukevs, line 20: does this hold it open? I was pretty sure this still dies immediately with `docker compose up` which is why the loop is required

lukevs: big fan if so, it's definitely better than a bash while loop

scripts/dev_tools.py
```diff
@@ -0,0 +1,31 @@
...
+    check_call(['black', '--diff', '--color', '.'])
+
+def start():
+    check_call(['docker', 'run', 'mev-inspect'])
+
+def build():
+    check_call(['docker', 'build', '-t', 'mev-inspect', '.'])
+
+def start_it():
+    check_call(['docker', 'run', 'mev-inspect'])
```
lukevs, line 31: can we change these to the docker compose equivalents? https://github.com/flashbots/mev-inspect-py/pull/29

lukevs:
Overall solid! Mind also updating the readme too 🙏

Dockerfile
pmdaly: The cmd line just runs that command when you enter, similar to specifying the same but through exec it. Added a tty arg to the compose solves the persistence.  Image below shows a few tests.  TTL is a clean solution.

Screen_Shot_2021-07-23_at_6 33 57_PM

scripts/dev_tools.py
pmdaly: using compose now

pmdaly:
Additional changes include using compose in poetry scripts, adding more docker commands, adjusting a few scripts to use cmdline args, no venv in docker - poetry install to base python, updating pyproject to pre-commit version of 3.9 and readme updates.

pmdaly:
updated the github actions and added cacheing to speed up build

lukevs (approved):

PR #31, opened 2021-07-23
Add alembic

lukevs:
## This PR

Alembic manages database migrations

To apply migrations up to the latest run:
```
poetry run exec alembic upgrade head
```

When we deploy the app for real we'll be able to use these to manage production DB changes as well

Migrations coming in next PRs

## Also

Makes some changes to poetry to:
- Fix commit hooks
- Add a generalized "exec" command for running alembic migrations and other things


pmdaly:
Is the additional requirements file necessary for Alembic? if we can keep all dependencies in pyproject.toml this will allow poetry to manage versioning and builds w/o conflicts.

Issue #11, opened 2021-07-07
Add Aave inspector


bertmiller:
As a reference you can use the MEV-Inspect-rs Aave inspector [here](https://github.com/flashbots/mev-inspect-rs/blob/master/src/inspectors/aave.rs)

PR #34, opened 2021-07-25
Adding Aave inspector

milancermak:
This PR adds Aave V2 liquidation inspector (#11).


milancermak:
I've opened this PR as a draft - I'd like to have some input on how to move forward. 

As is mentioned in #11, I've modeled it after the Rust inspector. I think I got the code to find a liquidation call right, but I'm unsure what to do next (it's my first inspector and there are no reducers yet). Looking at the Rust version, I probably need to modify `Classification` in `blocks.py` to have an `action` and a `trace_address`? (Note that in the initial commit, I created a placeholder `Liquidation` class, that now I realize it wrong).

If I understand the structure in the Rust version correctly, the idea is to have reducers independent of the inspectors right? In that case, just returning the correct classification from an inspector ought to be enough, is that so?

Note also I haven't tested the code in production yet, as I so far don't have an access to a full archive node. I've signed up for archivenode.io, fingers crossed they'll approve my request.

milancermak:
Closing since this PR is now outdated.

PR #37, opened 2021-07-27
Implement Trace Classifier and write to DB

lukevs:
## This PR

This PR is a bit large in that it does two big things:
- Replaces inspectors with the trace classifier + classifier specs
- Adds support for writing the classified traces to the DB

## Testing

Run an inspection like:
```
poetry run inspect --block-number 11931270 --rpc 'http://your-rpc'
```

Enter the DB (password is password for local)
```
docker compose exec db psql --username postgres --password -d mev_inspect
```

Check for entries
```
select count(*) from classified_traces where block_number = 11931270;
```


bertmiller:
Everything runs well on my end! Cheers

PR #38, opened 2021-07-28
Inspect args fix

pmdaly:
Block number and rpc are now correctly listed as separate arguments.  Added shorthands for optional args as well i.e., -b for --block-num


pmdaly:
had updated the inspect specific poetry script call but a more general pass through to docker script is preferred.

PR #41, opened 2021-07-29
Add ABIs and specs for Uniswap V3

lukevs:
## This PR

Adds V3 contracts from https://github.com/Uniswap/uniswap-v3-periphery/blob/main/deploys.md
+ the generic pool contract

Also moves ABIs into subfolders by protocol

We'll need to think about how to handle those non-protocol ones a little more

## Testing

**Run inspect for blocks with uniswap v3 action to see that it's working**

Block 12862188 uses the Quoter, SwapRouter, and V3 Pool
```
poetry run inspect --block-number 12862188 --rpc '...'

```

produces
```
{
    ...
    "UniswapV3Pool": {
        "swap": {
            "swap(address,bool,int256,uint160,bytes)": 6
        },
        "unknown": {
            "token0()": 6,
            "fee()": 1
        }
    },
    "SwapRouter": {
        "unknown": {
            "uniswapV3SwapCallback(int256,int256,bytes)": 2,
            "multicall(bytes[])": 1,
            "selfPermit(address,uint256,uint256,uint8,bytes32,bytes32)": 1
        }
    },
    "Quoter": {
        "unknown": {
            "quoteExactOutputSingle(address,address,uint24,uint256,uint160)": 1,
            "uniswapV3SwapCallback(int256,int256,bytes)": 1
        }
    }
}
```


mev_inspect/abi.py
```diff
@@ -5,14 +5,19 @@
...
 ABI_DIRECTORY_PATH = THIS_FILE_DIRECTORY / "abis"
 
 
-def get_abi(abi_name: str) -> Optional[ABI]:
-    abi_path = ABI_DIRECTORY_PATH / f"{abi_name}.json"
+def get_abi(abi_name: str, protocol: Optional[Protocol]) -> Optional[ABI]:
+    abi_filename = f"{abi_name}.json"
+    abi_path = (
+        ABI_DIRECTORY_PATH / abi_filename if protocol is None
+        else ABI_DIRECTORY_PATH / protocol.value / abi_filename
```
bertmiller, line 19: What's up here?

mev_inspect/classifier_specs.py
```diff
@@ -5,33 +5,109 @@
...
+    ),
+]
 
-CLASSIFIER_SPECS = [
+UNISWAPPY_V2_CONTRACT_SPECS = [
     ClassifierSpec(
         abi_name="UniswapV2Router",
         protocol=Protocol.uniswap_v2,
-        valid_contract_addresses=[UNISWAP_V2_ROUTER_ADDRESS],
+        valid_contract_addresses=["0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D"],
```
bertmiller, line 79: Why the change to a direct address instead of a variable?

mev_inspect/classifier_specs.py
```diff
@@ -5,33 +5,109 @@
...
+        "transfer(address,uint256)": Classification.transfer,
+        "burn(address)": Classification.burn,
+    },
+)
+
+
+CLASSIFIER_SPECS = (
+    UNISWAP_V3_CONTRACT_SPECS + 
+    UNISWAPPY_V2_CONTRACT_SPECS +
+    [ERC20_SPEC] +
```
bertmiller, line 110: why does this get wrapped in []?

bertmiller (changes requested):
Just a few small questions but otherwise LGTM!

mev_inspect/abi.py
lukevs: if there's a known protocol ABIs are listed at `abis/my_protocol/MyABI.json`

mev_inspect/classifier_specs.py
lukevs: Breaking the pattern of creating named constants for each.

Since this is their only use and they're functionally "named" as the valid contract address of that spec

mev_inspect/classifier_specs.py
lukevs: because the others are lists

lukevs: it does look a little gross though - I actually have a better way to write this

bertmiller (approved):
LGTM

PR #44, opened 2021-07-31
Curve trace

pmdaly:
First draft Curve specs.  Base Pools are spec'd out, next is Meta Pools and LiquidityGauge.   When specs are all merged into `CLASSIFIER_SPECS` at the end, there will be no distinction between a BasePool spec and a MetaPool spec other than the valid addresses.  Should we add classifications?


mev_inspect/abis/curve/DepositUSDP.json
```diff
@@ -0,0 +1 @@
...ame":"max_burn_amount"}],"stateMutability":"nonpayable","type":"function","gas":38200},{"name":"calc_withdraw_one_coin","outputs":[{"type":"uint256","name":""}],"inputs":[{"type":"uint256","name":"_token_amount"},{"type":"int128","name":"i"}],"stateMutability":"view","type":"function","gas":3147},{"name":"calc_token_amount","outputs":[{"type":"uint256","name":""}],"inputs":[{"type":"uint256[4]","name":"amounts"},{"type":"bool","name":"is_deposit"}],"stateMutability":"view","type":"function","gas":4414},{"name":"pool","outputs":[{"type":"address","name":""}],"inputs":[],"stateMutability":"view","type":"function","gas":1241},{"name":"token","outputs":[{"type":"address","name":""}],"inputs":[],"stateMutability":"view","type":"function","gas":1271},{"name":"base_pool","outputs":[{"type":"address","name":""}],"inputs":[],"stateMutability":"view","type":"function","gas":1301},{"name":"coins","outputs":[{"type":"address","name":""}],"inputs":[{"type":"uint256","name":"arg0"}],"stateMutability":"view","type":"function","gas":1440},{"name":"base_coins","outputs":[{"type":"address","name":""}],"inputs":[{"type":"uint256","name":"arg0"}],"stateMutability":"view","type":"function","gas":1470}]
```
lukevs, line 1: This looks like the same as DepositUSDN and some others. If so, we should just have a single Deposit ABI

mev_inspect/abis/curve/StableSwapUSDP.json
```diff
@@ -0,0 +1 @@
...name":"initial_A","outputs":[{"type":"uint256","name":""}],"inputs":[],"stateMutability":"view","type":"function","gas":2441},{"name":"future_A","outputs":[{"type":"uint256","name":""}],"inputs":[],"stateMutability":"view","type":"function","gas":2471},{"name":"initial_A_time","outputs":[{"type":"uint256","name":""}],"inputs":[],"stateMutability":"view","type":"function","gas":2501},{"name":"future_A_time","outputs":[{"type":"uint256","name":""}],"inputs":[],"stateMutability":"view","type":"function","gas":2531},{"name":"admin_actions_deadline","outputs":[{"type":"uint256","name":""}],"inputs":[],"stateMutability":"view","type":"function","gas":2561},{"name":"transfer_ownership_deadline","outputs":[{"type":"uint256","name":""}],"inputs":[],"stateMutability":"view","type":"function","gas":2591},{"name":"future_fee","outputs":[{"type":"uint256","name":""}],"inputs":[],"stateMutability":"view","type":"function","gas":2621},{"name":"future_admin_fee","outputs":[{"type":"uint256","name":""}],"inputs":[],"stateMutability":"view","type":"function","gas":2651},{"name":"future_owner","outputs":[{"type":"address","name":""}],"inputs":[],"stateMutability":"view","type":"function","gas":2681}]
```
lukevs, line 1: Same as Deposit. If they're the same just a single StableSwap.json

mev_inspect/classifiers/specs/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .uniswap import UNISWAP_CLASSIFIER_SPECS
+from .curve import CURVE_CLASSIFIER_SPECS
+
+ALL_CLASSIFIER_SPECS = UNISWAP_CLASSIFIER_SPECS + CURVE_CLASSIFIER_SPECS
```
lukevs, line 4: love it

mev_inspect/classifiers/specs/curve.py
```diff
@@ -0,0 +1,440 @@
...
+        protocol=Protocol.curve,
+        valid_contract_addresses=["0xeB21209ae4C2c9FF2a86ACA31E123764A3B6Bc06"],
+    ),
+    ClassifierSpec(
+        abi_name="DepositPAX",
+        protocol=Protocol.curve,
+        valid_contract_addresses=["0xA50cCc70b6a011CffDdf45057E39679379187287"],
+    ),
+    ClassifierSpec(
+        abi_name="DepositsUSD",
```
lukevs, line 187: same comment as before that these can be a single ABI

Might be worth thinking about how we can put more metadata like that into these things, but for the purpose of arbs at least we'll be able to know the asset from the transfers inside

pmdaly: I think the DepositsUSD is a staked usd deposit which is different than the other USDN/K/P contracts.

mev_inspect/abis/curve/DepositUSDP.json
pmdaly: merged the N/K/P contracts into a single DepositUSD contract

mev_inspect/abis/curve/StableSwapUSDP.json
pmdaly: Merged usdk and usdn into a single contract, usdp has around 50 differences

mev_inspect/block.py
```diff
@@ -51,7 +51,7 @@ def fetch_block(w3, base_provider, block_number: int) -> Block:
 
     ## Get gas used by individual txs and store them too
     txs_gas_data: Dict[str, Dict[str, Any]] = {}
-
+    """
```
lukevs, line 54: meant to be here still?

we should probably delete this code for now if it's not being used since it causes block loading to be slow

mev_inspect/classifiers/specs/uniswap.py
```diff
@@ -118,7 +118,7 @@
 )
 
 
-CLASSIFIER_SPECS = [
+UNISWAP_CLASSIFIER_SPECS = [
```
lukevs, line 121: AAVE_SPEC, ERC20_SPEC, and WETH_SPEC should be in their own files instead of the uniswap one

mev_inspect/block.py
pmdaly: Wasn't clear on if it were to be added back in.  Removed for the time being.

mev_inspect/classifiers/specs/uniswap.py
pmdaly: Done.  ERC20 must be the first in the ALL_CLASSIFIER_SPECS list or arbs won't be caught in testing.

lukevs (approved):

PR #46, opened 2021-08-05
Find arbitrages in a block

lukevs:
## This PR
Sorry this is a big PR - lots of the lines are tests!

Adds support for finding arbitrage in UniswappyV2 and UniswapV3 swaps

Also:
- Adds support for weth
- Adds structured objects for a Transfer and Swap and functions for building them from traces
- Removes the unused NestedTrace code (at least for now)

## Reading Roadmap

I recommited all the code in reading order, so click the first one then walk through in order

## Testing

Verified it finds the same arb as MevAlphaLeak in block 12914944
```
poetry run inspect --block-number 12914944 --rpc '...'
```

Also added some unit tests for each piece


mev_inspect/transfers.py
```diff
@@ -0,0 +1,62 @@
...
+
+        if from_address is not None and transfer.from_address != from_address:
+            continue
+
+        filtered_transfers.append(transfer)
+
+    return filtered_transfers
+
+
+def remove_inner_transfers(transfers: List[Transfer]) -> List[Transfer]:
```
taarushv, line 41: Can you share a bit about `inner_transfers` int his context? I see usage in `_get_swaps_for_transaction` and `test_remove_inner_transfers` (which returns as expected) but still not sure I follow what's going on here wrt swaps.

mev_inspect/schemas/arbitrage.py
```diff
@@ -0,0 +1,14 @@
...
+from .swaps import Swap
+
+
+class Arbitrage(BaseModel):
+    swaps: List[Swap]
+    account_address: str
+    profit_token_address: str
+    start_amount: int
+    end_amount: int
+    profit_amount: int
```
taarushv, line 14: I see `profit_amount` being assigned `end_amount - start_amount` in the impl below, do we want to keep the amount denominated in the token the arb was performed in or ETH (at the time)? previous inspect impl converted all profits to ETH for consistency but I'm open to having `profit_amount` and `profit_amount_eth` to help us with advanced queries/debugging etc.

mev_inspect/schemas/swaps.py
```diff
@@ -0,0 +1,20 @@
+from typing import List, Optional
+
+from pydantic import BaseModel
+
+from mev_inspect.schemas.classified_traces import Protocol
+
+
+class Swap(BaseModel):
```
taarushv, line 8: What do you think of an additional property like `slippage_tolerance` that'd calculate it based on historical reserves (when the tx was included). Default is 0.5% on uniswap but I've a feeling it'd come in handy when we implement sandwich detection or just want to get a better idea.

tests/test_arbitrages.py
```diff
@@ -0,0 +1,160 @@
+from mev_inspect.arbitrage import get_arbitrages
+from mev_inspect.schemas.swaps import Swap
+from mev_inspect.swaps import (
+    UNISWAP_V2_PAIR_ABI_NAME,
+    UNISWAP_V3_POOL_ABI_NAME,
+)
+
+
+def test_two_pool_arbitrage(get_transaction_hashes, get_addresses):
+    block_number = 123
```
taarushv, line 10: Def a neat way to construct fixtures/tests, but can we use the same test files to plug in real tx hashes and see expected arb profit output?

taarushv (commented):
Added comments for where I need clarification but looks great overall!

mev_inspect/schemas/arbitrage.py
lukevs: Yeah I think separating `profit_amount` from `profit_amount_eth` is good. Basically trying to provide the most un-opinionated data here. ETH price at the time feels like an interpretation on top.

Like maybe we have a separate function prior to the DB which adds that information.

mev_inspect/schemas/swaps.py
lukevs: Is that included as a parameter of the swap usually? If so I'm definitely for it.

If it requires some additional lookups / calculation, I think it'd be better as an added attribute later to keep this object to the data that's derivable directly from the trace (similar argument for separating `profit_amount_eth` on `Arbitrage`, keep it purely un-opinionated at this level)

mev_inspect/traces.py
```diff
@@ -1,80 +1,33 @@
...
-from typing import Iterable, List
+from typing import List
 
-from mev_inspect.schemas import Trace, NestedTrace
+from mev_inspect.schemas.classified_traces import ClassifiedTrace
 
 
-def as_nested_traces(traces: Iterable[Trace]) -> List[NestedTrace]:
-    nested_traces = []
+def is_child_trace_address(
```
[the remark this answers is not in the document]
lukevs: Yeah that's fair to say. Nested trace object basically just isn't used. I've at least found it easier to look at the traces linearly rather than reconstructing the tree. Might be a future use case where we want to bring the tree back though.

mev_inspect/transfers.py
lukevs: This is a good call out. I was struggling to find a good name for this

Basically I noticed that USDC specifically internally has another ERC-20 that it calls (FiatToken iirc). So you'll see an ERC-20 transfer with a child trace of another ERC-20 transfer of all the same parameters but a separate token address.

For the swaps, it's easier to ignore that implementation detail and just say that we swapped the parent token (USDC). So this removed any transfers which are children of other transfers

Maybe `remove_child_transfers_of_transfers`?

tests/test_arbitrages.py
lukevs: We could add some tests using real transfers by saving the JSON in the tests. I'll add that

Otherwise I did some manual tests for blocks with arbs I found people post about on twitter.

mev_inspect/schemas/swaps.py
taarushv: One of the param's expresses it (ex: `AmountOutMin` in `swapExactTokensForETH`) but you'd still need to query historical pool reserves to get the exact %. As you said, it makes sense to keep this un-opinionated so we can do this pre-db insert.

mev_inspect/transfers.py
taarushv: Makes sense, better renaming (or just adding code comments for context around USDC-like tokens) should be good. The rest of the PR makes sense too so should be good to merge post this fix.

taarushv (approved):

PR #47, opened 2021-08-11
Liquidation inspector

gheise:
The liquidation inspector as of now:

- Identifies and returns AAVE liquidations
- Identifies the liquidator's transfers, which is not returned
- Once profit calculation is handled, LiquidationData will be returned instead with the transfers and the profit
- Unit tests not yet written, manual testing only has been done, but this will be the next implementation


liquidation_inspector.py
```diff
@@ -0,0 +1,105 @@
...
+import pandas as pd
+from typing import Optional
+from web3 import Web3
+from typing import List, Optional
+
+# TODO: adjust profit to new model
+#        unit test
+#        coinbase check / collateral source
+
+liquidations = []
```
lukevs, line 22: any non-constant values should be inside a function

global variables in Python are difficult to reason about - it's like sneaky passing a variable into all functions since anyone can access and modify it. Makes it difficult to know what does and doesn't depend on it, and how the values changed

They also make it difficult to scale code - running two processes / instances of this code will create two separate global variables while the code itself might still be assuming it's one shared one.

liquidation_inspector.py
```diff
@@ -0,0 +1,105 @@
...
+            trace.classification == Classification.liquidate and
+            trace.transaction_hash not in tx
+            ):
+
+            liquidations.append(trace)
+            addrs.append(trace.from_address)
+            tx.append(trace.transaction_hash)
+
+            # Found liquidation, now parse inputs for data
+            for i in trace.inputs:
```
lukevs, line 49: `i` implies a counter like `for i in range(10)` - what is this really?

liquidation_inspector.py
```diff
@@ -0,0 +1,105 @@
...
+                if(i == '_purchaseAmount'):
+                    liquidation_amount = trace.inputs[i]
+                elif (i == '_collateral'):
+                    collateral_type = trace.inputs[i]
+                    #This will be the address of the sent token
+                elif (i == '_reserve'):
+                    reserve = trace.inputs[i]
+                    #This will be the address of the received token
+                elif(i == '_user'):
+                    liquidated_usr = trace.inputs[i]
```
lukevs, line 60: these variables are each created inside an `if` branch which implies it's ok for them not to exist / they only are used inside of the `if`

but if the branch isn't hit for all of them the `Liquidation` creation will fail with a `No such variable found "liquidated_usr"` for example

lukevs: Also why `usr` not `user`?

One of my favorite articles in all of programming:
https://javascript.info/ninja-code

Take 5 mins and read it (ignore that it's about javascript - they're universal lessons)

Here's what they would have to say
image

liquidation_inspector.py
```diff
@@ -0,0 +1,105 @@
...
+            # Register liquidation
+            result.append(
+                            Liquidation(collateral_type=collateral_type,
+                                            collateral_amount=liquidation_amount,
+                                            collateral_source="",
```
lukevs, line 65: Better to make it `None` to allow a consumer of the data to say:
```
if liquidation.collateral_source is not None:
    valid_source = liquidation.collateral_source
```

liquidation_inspector.py
```diff
@@ -0,0 +1,105 @@
...
+        elif (
+              trace.classification == Classification.transfer and
+              trace.inputs['recipient'] in addrs
+             ):
+             #Add the transfer
+             liquidator = next(addrs[i] for i in range(len(addrs)) if trace.inputs['recipient'] == addrs[i])
+             transfers_to.append(["to", liquidator, trace.transaction_hash, trace.inputs['amount']])
+
+    for count, trace in enumerate(liquidations):
+        tx = trace.transaction_hash
```
lukevs, line 94: `tx` was a list but now it is a string

mev_inspect/schemas/strategy.py
```diff
@@ -0,0 +1,22 @@
+from typing import Optional, List
+from enum import Enum
+from pydantic import BaseModel
+from .classified_traces import ClassifiedTrace, Protocol
+
+class StrategyType(Enum):
+    arbitrage = "arbitrage"
+    liquidation = "liquidation"
+
+class Strategy(BaseModel):
```
lukevs, line 10: I think better for now to just have a `Liquidation` without a base `Strategy` object.

We can combine them if we find a use for something being either a `Liquidation` or an `Arbitrage`

liquidation_inspector.py
```diff
@@ -0,0 +1,105 @@
...
+             transfers_from.append(["from", liquidator, trace.transaction_hash, trace.inputs['amount']])
+             from_doubles.append(trace.transaction_hash)
+
+        # Check for transfer to a liquidator
+        elif (
+              trace.classification == Classification.transfer and
+              trace.inputs['recipient'] in addrs
+             ):
+             #Add the transfer
+             liquidator = next(addrs[i] for i in range(len(addrs)) if trace.inputs['recipient'] == addrs[i])
```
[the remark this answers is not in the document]
gheise: not the same as addrs[0]; This will return the address of the liquidator if the recipient is present in addrs;
You can think next as a function returning the 0th index of a list comprehension which is the argument

PR #52, opened 2021-08-18
Call Interface

gheise:
Separates classified traces by attributes:

1.  ClassifiedTraceStrict(BaseModel) Baseline, transaction hashes, block number, etc.
2. ClassifiedCall(ClassifiedTraceStrict): Adds recipient/sender addresses, amount sent, and gas
3. ClassifiedCallData(ClassifiedCall): Adds protocol, function tags, inputs, and ABI

The overall goal is to separate the traces as early as possible to facilitate handling them in the inspectors and summarizers. This removes all but one optional paramater, which removes " if x " checks throughout the codebase, removes ambiguity about the shape of the trace, and still supports current implementation of traces in the inspectors.

Note on naming: The presence of 'Strict' will be irrelevant once this is fully implemented, but for now it allows for the usage of both objects

### DIFF
--- a/mev_inspect/schemas/classified_traces.py
+++ b/mev_inspect/schemas/classified_traces.py
@@ -63,7 +63,7 @@ class ClassifiedTraceStrict(BaseModel):
         }
 
 
-class ClassifiedCall(ClassifiedTrace):
+class ClassifiedCall(ClassifiedTraceStrict):
     to_address: str
     from_address: str
     gas: int
--- a/mev_inspect/schemas/classified_traces.py
+++ b/mev_inspect/schemas/classified_traces.py
@@ -28,6 +28,32 @@ class ClassifiedTrace(BaseModel):
     trace_type: TraceType
     trace_address: List[int]
     classification: Classification
+    protocol: Optional[Protocol]
+    abi_name: Optional[str]
+    function_name: Optional[str]
+    function_signature: Optional[str]
+    inputs: Optional[Dict[str, Any]]
+    to_address: Optional[str]
+    from_address: Optional[str]
+    gas: Optional[int]
+    value: Optional[int]
+    gas_used: Optional[int]
+    error: Optional[str]
+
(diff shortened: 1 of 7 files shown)


mev_inspect/schemas/transfers.py
```diff
@@ -14,7 +17,7 @@ class Transfer(BaseModel):
     token_address: str
 
     @classmethod
-    def from_trace(cls, trace: ClassifiedTrace) -> "Transfer":
+    def from_trace(cls, trace) -> "Transfer":
```
lukevs, line 20: why no types?

mev_inspect/swaps.py
```diff
@@ -63,7 +64,7 @@ def _get_swaps_for_transaction(traces: List[ClassifiedTrace]) -> List[Swap]:
 
 
 def _parse_swap(
-    trace: ClassifiedTrace,
+    trace,
```
lukevs, line 67: no types?

If this creates an issue with mypy because there are potentially two types getting passed in you can use the `Union` type

`Union[ClassifiedTrace, ClassifiedTraceStrict]`

mev_inspect/schemas/classified_traces.py
```diff
@@ -48,6 +48,37 @@ class Config:
...
+    gas_used: Optional[int]
+
+
+class ClassifiedCallData(ClassifiedCall):
+    protocol: Optional[Protocol]
+    function_name: str
+    function_signature: str
+    inputs: Dict[str, Any]
+    abi_name: str
+
```
lukevs, line 80: Since these are all still "Traces" I think that should still be the main noun for the object

ex:
```
StrictClassifiedTrace
ClassifiedCallTrace
ClassifiedCallDataTrace
```

Another option, reserve classified for when we actually knew what the ABI was. So it'd be
```
Trace
CallTrace
ClassifiedCallTrace
```

PR #63, opened 2021-09-03
Balancer Contracts

pmdaly:
Currently all balancer transactions are being classified as ERC20 or Vault despite the addresses and abi's being unique.  Coult use a second set of eyes on this.


pmdaly:
Addresses found https://docs.balancer.fi/developers/smart-contracts/deployment-addresses 
ABIs / Addresses also found https://github.com/balancer-labs/balancer-v2-monorepo/tree/master/pkg/deployments/tasks

lukevs:
> Currently all balancer transactions are being classified as ERC20 or Vault despite the addresses and abi's being unique. Coult use a second set of eyes on this.

Do the other ABIs inherit from ERC20 / Vault? That would explain them all getting classified as Vault since it's the highest up

This is the same issue we've had with other contracts that inherit from each other. I think whether it should be classified as the more generic or the more specific is kind of a case by case issue, but so far we've gone for the more generic

PR #64, opened 2021-09-09
Add support for running mev-inspect on local kubernetes

lukevs:
## What's this
Previously mev-inspect has been run using docker compose

For production deployment we'll be using kubernetes, and it's ideal to have our local and production environments as similar as possible

This keeps all of the existing docker compose pieces in place so people can switch over gracefully, but they'll be removed later

### Install the kubernetes control tool `kubectl`

Kubectl (pronunciation [debatable](https://www.youtube.com/watch?v=2wgAIvXpJqU&ab_channel=Opensource.com)) is the main CLI tool for talking to kubernetes (applying updates, checking whether something is running, etc)

Install: https://kubernetes.io/docs/tasks/tools/

### Set up a local kube cluster
Install a local kube cluster - I like [kind](https://kind.sigs.k8s.io/) because it's small and fast

Installing kind: https://kind.sigs.k8s.io/docs/user/quick-start/#installation

If using kind, you can create the default cluster with
```
kind create cluster
```

### Install Tilt

Tilt manages creation / updates for local dev kubernetes (similar to `docker-compose up`)

Install: https://docs.tilt.dev/install.html

### Install Helm

Helm is like pip for kubernetes - right now we use it for setting up a postgres DB
Install: https://helm.sh/docs/intro/install/

### Stop the running docker compose setup
```
poetry run stop
```

### Start the kube setup
Verify your `kubectl` is talking to your local kube instance by running
```
kubectl config current-context
```
For kind, you should see `kind-kind`

Start the kube setup with
```
tilt up
```

From the terminal where tilt is started, press "space" to see a browser view of all systems coming live

### Apply DB migrations

This will spin up a new DB instance, so you'll need to re-run migrations on it

```
kubectl exec deploy/mev-inspect-deployment -- alembic upgrade head
```

### Test on a block
Then run inspect on a block (ex: 12914944) with
```
kubectl exec deploy/mev-inspect-deployment -- poetry run inspect-block 12914944 'http://your-rpc'
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@ htmlcov
 
 # don't commit cache
 cache
+
+# k8s
+.helm
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,14 +6,16 @@ RUN pip install -U pip \
 
 ENV PATH="${PATH}:/root/.poetry/bin"
 
-COPY . /app
+COPY ./pyproject.toml /app/pyproject.toml
+COPY ./poetry.lock /app/poetry.lock
 WORKDIR /app/
 
-# poetry uses virtual env by default, turn this off inside container
 RUN poetry config virtualenvs.create false && \
     poetry install
+
+COPY . /app
 
 # easter eggs 😝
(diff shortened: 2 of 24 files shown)


bertmiller (approved):
LGTM

PR #66, opened 2021-09-14
Poll for incoming blocks and auto-inspect them

lukevs:
## What's this

- Adds a script `run.py` which on start of the container starts polling for blocks it hasn't processed yet and inspects them
- breaks inspect_block's logic from the script into its own module
- Checkpoints the latest processed block in the `latest_block_update` table

## Reading Roadmap

You can ignore the `tilt_modules` - those are just locked dependencies

Start with `run.py` and work your way down


mev_inspect/block.py
```diff
@@ -70,6 +74,8 @@ def get_transaction_hashes(calls: List[Trace]) -> List[str]:
 def cache_block(cache_path: Path, block: Block):
     write_mode = "w" if cache_path.is_file() else "x"
 
+    cache_path.parent.mkdir(parents=True, exist_ok=True)
```
lukevs, line 77: create the cache dir if it doesn't exist

bertmiller:
Getting some merge conflicts here!

mev_inspect/inspect_block.py
```diff
@@ -0,0 +1,76 @@
+import logging
```
lukevs, line 1: moved from scripts/inspect_block.py

PR #67, opened 2021-09-14
Add block number index on miner_payments

lukevs:
On each inspect we delete all miner payments for a block

We also query a lot by block for miner payments

This adds an index to make that query faster


lukevs:
making a new one

PR #69, opened 2021-09-14
Make mev-inspect require that the DB is already up for local dev

lukevs:
Otherwise it can start and die and need restart


bertmiller:
Conflict on tiltfile!

PR #70, opened 2021-09-16
Add stricter types to trace classifier

gheise:
This is an adapted version of the draft PR regarding stricter types. Taking in some of the feedback on the directory structure, naming, and database writing, there has been some changes:

- Directory has been kept the same until further action is needed rather
- Naming now emphasizes Trace property rather than arbitrary object name
- Object attributes can now be filtered manually before written to the database, which is useful as some attrs can
be used during classification and strategy but can be discarded before db writing is done
- Serialization of dictionaries containing byte and tuple of bytes, like trace.inputs, can now be done separately

And some other minor updates


mev_inspect/crud/classified_traces.py
```diff
@@ -22,9 +22,30 @@ def write_classified_traces(
...
     classified_traces: List[ClassifiedTrace],
 ) -> None:
-    models = [
-        ClassifiedTraceModel(**json.loads(trace.json())) for trace in classified_traces
-    ]
+    models = []
+    for trace in classified_traces:
+        models.append(
+            ClassifiedTraceModel(
+                **{
```
lukevs, line 29: the spread (**) is for cases where you have a dict and want to assign the keys to the args of a function.

Since you explicitly know the argument names, better to do 
```
ClassifiedTraceModel(
    transaction_hash=trace.transaction_hash,
    block_number=trace.block_number
    ...
```

mev_inspect/schemas/classified_traces.py
```diff
@@ -24,32 +24,53 @@ class Protocol(Enum):
...
     to_address: Optional[str]
     from_address: Optional[str]
     gas: Optional[int]
     value: Optional[int]
     gas_used: Optional[int]
-    error: Optional[str]
+    protocol: Optional[Protocol]
+    function_name: Optional[str]
+    function_signature: Optional[str]
+    inputs: Optional[Dict[str, Any]]
```
lukevs, line 44: is it `Any`? I thought you learned it was bytes always

mev_inspect/schemas/classified_traces.py
```diff
@@ -24,32 +24,53 @@ class Protocol(Enum):
...
+        validate_assignment = True
         json_encoders = {
             # a little lazy but fine for now
             # this is used for bytes value inputs
             bytes: lambda b: b.hex(),
         }
 
 
+class CallTrace(ClassifiedTrace):
+
```
lukevs, line 57: no need for this space

mev_inspect/schemas/classified_traces.py
```diff
@@ -24,32 +24,53 @@ class Protocol(Enum):
...
+    to_address: str
+    from_address: str
+
+
+class DecodedCallTrace(CallTrace):
+    inputs: Dict[str, Any]
+    abi_name: str
+
+    # Optional
```
lukevs, line 66: I think remove this comment

Controversial opinion, but in general I'm not a fan of comments because they're out of sync with the code itself (ppl can change code without remembering to change comments => more confusion / issues - do I trust the comment or my own interpretation of the code?). Better to just write clear code.

Example for this one:
- This comment is here
- Someone doesn't see the comment and puts an optional in the top ones (now the comment is wrong)
- Someone else sees the comment and writes their code assuming that those are the only Optional arguments
- Later they get an error in their code because they missed one someone put up top

I think just slotting them in the bottom is enough to imply it without making some guarantee

mev_inspect/schemas/classified_traces.py
gheise: It can actually also be a tuple of bytes

mev_inspect/schemas/utils.py
```diff
@@ -15,6 +15,25 @@ def to_original_json_dict(model: BaseModel) -> dict:
     return json.loads(model.json(by_alias=True, exclude_unset=True))
 
 
+def to_json(d):
```
lukevs, line 18: single letter variables are code ninja shit

mev_inspect/schemas/utils.py
```diff
@@ -15,6 +15,25 @@ def to_original_json_dict(model: BaseModel) -> dict:
...
+    if isinstance(d, dict):
+        tmp = {}
+        for key, value in d.items():
+            if isinstance(value, bytes):
+                tmp[key] = value.hex()
+            if isinstance(value, tuple):
+                res = []
```
lukevs, line 25: use full words. characters are free. short names are expensive for the reader.

tests/helpers.py
```diff
@@ -12,11 +17,13 @@ def make_transfer_trace(
     to_address: str,
     token_address: str,
     amount: int,
+    action: dict = {},
```
lukevs, line 20: Scary python thing here

Using a mutable type (ex: dict, list) as a default shares the same memory across all default usages!

Example:
```
def myfunc(x = []):
    x.append(1)
    print(x)

myfunc()
myfunc()
```

prints
```
[1]
[1, 1]
```

The common pattern is to instead do:
```
def myfunc(maybe_x = None):
    x = [] if maybe_x is None else maybe_x
    x.append(1)
    print(x)
```

which now gives
```
[1]
[1]
```

tests/helpers.py
```diff
@@ -12,11 +17,13 @@ def make_transfer_trace(
...
+    subtraces: int = 0,
```
lukevs, line 21: rather than giving people the _option_ to pass these, better to make it _explicit_ that no one does and not have them as arguments. Just set the values directly where they're used inside the function.

mev_inspect/crud/classified_traces.py
```diff
@@ -22,9 +22,30 @@ def write_classified_traces(
...
+                    "transaction_hash": trace.transaction_hash,
+                    "block_number": trace.block_number,
+                    "classification": trace.classification.value,
+                    "trace_type": trace.type.value,
+                    "trace_address": trace.trace_address,
+                    "protocol": str(trace.protocol),
+                    "abi_name": trace.abi_name,
+                    "function_name": trace.function_name,
+                    "function_signature": trace.function_signature,
+                    "inputs": json.loads(trace.json(include={"inputs"}))["inputs"],
```
lukevs, line 39: this is a heavy line. I think better to pop it out like

```
inputs_json = json.loads(trace.json(include={"inputs"}))["inputs"]
```
then use it inline

PR #72, opened 2021-09-17
Write transfers to transfers table on inspect

lukevs:
Gives us the ability to query for all token flows in each block


bertmiller (approved):
LGTM

PR #74, opened 2021-09-17
Add AAVE liquidations to inspect_block

gheise:
Adds AAVE liquidations to inspect block

1. Returns identified liquidations from inspector and prints on inspect block without writing to DB
2. Database writing and profit/collateral source logic will come in next separate PR


mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,113 @@
+from typing import List
+
+from mev_inspect.schemas.classified_traces import (
+    Classification,
+    ClassifiedTrace,
+    # CallTrace,
```
lukevs, line 6: should remove these if they’re not used

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,113 @@
...
+    DecodedCallTrace,
+)
+from mev_inspect.schemas.liquidations import Liquidation
+
+
+# Inspect list of classified traces and identify liquidation
```
lukevs, line 12: in python this sort of comment should be a docstring - example: https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,113 @@
...
+def get_liquidations(traces: List[ClassifiedTrace]):
```
lukevs, line 13: return type

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,113 @@
...
+    tx = []
+    liquidations = []
+    result = []
+
+    # Protocol contract address must be in included, below is AaveLendingPoolCoreV1
+    addrs = [
```
lukevs, line 19: if these are constant, throw them up top in a constant

also no short names

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,113 @@
...
+        "0x7d2768dE32b0b80b7a3454c06BdAc94A69DDc7A9",
+        "0x398eC7346DcD622eDc5ae82352F02bE94C62d119",
+    ]
+
+    # Used to remove double-counted 'from' transfers
+    from_doubles = []
+    transfers_to = []
+    transfers_from = []
+
+    # For each trace
```
lukevs, line 30: this comment isn’t needed

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,113 @@
...
+    for trace in traces:
+
+        if isinstance(trace, DecodedCallTrace):
+
+            # Check for liquidation and register trace and unique liquidator
```
lukevs, line 35: rather than having this be a comment, better to break up the code so it reads the same as the comment would - break this inner piece out to a named function

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,113 @@
...
+
+    # Protocol contract address must be in included, below is AaveLendingPoolCoreV1
+    addrs = [
+        "0x3dfd23A6c5E8BbcFc9581d2E864a68feb6a076d3",
+        "0x7d2768dE32b0b80b7a3454c06BdAc94A69DDc7A9",
+        "0x398eC7346DcD622eDc5ae82352F02bE94C62d119",
+    ]
+
+    # Used to remove double-counted 'from' transfers
+    from_doubles = []
```
lukevs, line 26: there’s probably a better name for this that doesn’t require a comment

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,113 @@
...
+                    elif input == "_collateral":
+                        collateral_address = trace.inputs[input]
+                        # This will be the address of the sent token
+                    elif input == "_reserve":
+                        reserve = trace.inputs[input]
+                        # This will be the address of the received token
+                    elif input == "_user":
+                        liquidated_usr = trace.inputs[input]
+                # Register liquidation
+                result.append(
```
lukevs, line 59: based on how the loop is written, if you never get a “_user” input, you’ll get a variable not defined error for liquidated usr here (same for the other variables)

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,113 @@
...
+                    if input == "_purchaseAmount":
+                        liquidation_amount = trace.inputs[input]
+                    elif input == "_collateral":
+                        collateral_address = trace.inputs[input]
+                        # This will be the address of the sent token
+                    elif input == "_reserve":
+                        reserve = trace.inputs[input]
+                        # This will be the address of the received token
+                    elif input == "_user":
+                        liquidated_usr = trace.inputs[input]
```
lukevs, line 57: usr => user. characters are free. reading takes time and energy

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,113 @@
...
+                        liquidated_usr=liquidated_usr,
+                        collateral_address=collateral_address,
+                        collateral_amount=liquidation_amount,
+                        collateral_source="",
+                        reserve=reserve,
+                    )
+                )
+
+            # Check for transfer from a liquidator
+            elif (
```
lukevs, line 70: instead of a comment, make a named function

ex:
```
if is_transfer_from_liquidator(…)
```

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,113 @@
...
+                )
+                from_doubles.append(trace.transaction_hash)
+
+            # Check for transfer to a liquidator
+            elif (
+                trace.classification == Classification.transfer
+                and trace.inputs["recipient"] in addrs
+            ):
+                # Add the transfer
+                liquidator = next(
```
lukevs, line 94: if the number of matching addrs is 0, this raises a StopIteration exception

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,113 @@
...
+                    addrs[i]
+                    for i in range(len(addrs))
+                    if trace.inputs["recipient"] == addrs[i]
+                )
+                transfers_to.append(
```
lukevs, line 99: what is the type of transfers_to?

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,113 @@
...
+                    ["to", liquidator, trace.transaction_hash, trace.inputs["amount"]]
+                )
+
+    return result
+
+    # for count, trace in enumerate(liquidations):
```
lukevs, line 105: if unused remove

mev_inspect/schemas/liquidations.py
```diff
@@ -0,0 +1,16 @@
...
+
+class Liquidation(BaseModel):
+    liquidated_usr: str
+    collateral_address: str
+    collateral_amount: int
+    collateral_source: str
+    reserve: str
+
+
+class LiquidationData(Liquidation):
```
lukevs, line 14: is this used?

scripts/inspect_block.py
```diff
@@ -120,6 +122,14 @@ def _inspect_block(
         delete_arbitrages_for_block(db_session, block_number)
         write_arbitrages(db_session, arbitrages)
 
+    liquidations = get_liquidations(classified_traces)
+    click.echo(f"Found {len(liquidations)} liquidations")
+    print(liquidations)
+
+    # if should_write_liquidations:
```
lukevs, line 129: if unused remove

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,113 @@
...
+                # Add the transfer
+                liquidator = next(
+                    addrs[i]
+                    for i in range(len(addrs))
+                    if trace.inputs["sender"] == addrs[i]
+                )
+                transfers_from.append(
+                    ["from", liquidator, trace.transaction_hash, trace.inputs["amount"]]
+                )
+                from_doubles.append(trace.transaction_hash)
```
lukevs, line 86: `seen_transactions` instead?

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,113 @@
...
+                        collateral_amount=liquidation_amount,
+                        collateral_source="",
+                        reserve=reserve,
+                    )
+                )
+
+            # Check for transfer from a liquidator
+            elif (
+                trace.classification == Classification.transfer
+                and "sender" in trace.inputs
```
lukevs, line 72: sender is not always the from. sometimes it’s just from address of the trace. 

instead of doing this logic here, it’s already done as part of the Transfers objects

try this:
```
from mev_inspect.schemas.transfers import Transfer

if trace.classification == Classification.transfer:
    transfer = Transfer.from_trace(trace)
    transfer.from_address
```

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,113 @@
...
+                and trace.inputs["sender"] in addrs
+                and trace.transaction_hash not in from_doubles
+            ):
+
+                # Add the transfer
+                liquidator = next(
```
lukevs, line 78: same point as below about StopIteration

mev_inspect/schemas/liquidations.py
```diff
@@ -0,0 +1,16 @@
+from typing import List
+from pydantic import BaseModel
+from .classified_traces import ClassifiedTrace
+
+
+class Liquidation(BaseModel):
+    liquidated_usr: str
```
lukevs, line 7: usr => user

mev_inspect/schemas/liquidations.py
```diff
@@ -0,0 +1,16 @@
...
+    collateral_address: str
+    collateral_amount: int
+    collateral_source: str
```
lukevs, line 10: what is collateral source? is it the token used as collateral?

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,113 @@
...
+    ClassifiedTrace,
+    # CallTrace,
+    DecodedCallTrace,
+)
+from mev_inspect.schemas.liquidations import Liquidation
+
+
+# Inspect list of classified traces and identify liquidation
+def get_liquidations(traces: List[ClassifiedTrace]):
+    tx = []
```
lukevs, line 14: better name - what is this used for?

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,113 @@
...
+    addrs = [
+        "0x3dfd23A6c5E8BbcFc9581d2E864a68feb6a076d3",
+        "0x7d2768dE32b0b80b7a3454c06BdAc94A69DDc7A9",
+        "0x398eC7346DcD622eDc5ae82352F02bE94C62d119",
+    ]
+
+    # Used to remove double-counted 'from' transfers
+    from_doubles = []
+    transfers_to = []
+    transfers_from = []
```
lukevs, line 28: it doesn't look like these are ever accessed

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,128 @@
...
+
+from mev_inspect.schemas.classified_traces import (
+    ClassifiedTrace,
+    DecodedCallTrace,
+    Classification,
+)
+from mev_inspect.schemas.liquidations import Liquidation
+from mev_inspect.schemas.transfers import ERC20Transfer
+
+contract_addresses = [
```
lukevs, line 11: which contract addresses? what do these represent?

also for constants the convention is capitals `CONTRACT_ADDRESSES`

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,128 @@
...
+    unique_transfer_hashes: List[str],
+) -> bool:
+    """Check if transfer is from liquidator"""
+    transfer = ERC20Transfer.from_trace(trace)
+    if (
+        trace.classification == Classification.transfer
+        and transfer.from_address in contract_addresses
+        and trace.transaction_hash not in unique_transfer_hashes
+    ):
+        return True
```
lukevs, line 114: For both of these, any time you have
```
if x:
    return True
else:
    return False
```

it's the same as
```
return x
```

(same thing in reverse with a not)

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,128 @@
...
+                )
+
+    return [result, transfers_to, transfers_from]
+
+
+def is_transfer_from_liquidator(
+    trace: ClassifiedTrace,
+    unique_transfer_hashes: List[str],
+) -> bool:
+    """Check if transfer is from liquidator"""
```
lukevs, line 107: no need for this - it's the same as the function name

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,128 @@
...
+                        for i in range(len(contract_addresses))
+                        if trace.inputs["recipient"] == contract_addresses[i]
+                    ),
+                    "",
+                )
+                transfers_to.append(
+                    ["to", liquidator, trace.transaction_hash, trace.inputs["amount"]]
+                )
+
+    return [result, transfers_to, transfers_from]
```
lukevs, line 100: are your pre-commit hooks installed? this doesn't look like it matches the return type

gheise: Yes, this does not conflict with the List return type

mev_inspect/aave_liquidations.py
gheise: Done

mev_inspect/aave_liquidations.py
gheise: Done

mev_inspect/aave_liquidations.py
gheise: How can there not be a _user input?

mev_inspect/aave_liquidations.py
gheise: done

mev_inspect/aave_liquidations.py
gheise: done

mev_inspect/aave_liquidations.py
gheise: done

mev_inspect/schemas/liquidations.py
gheise: No, removed

scripts/inspect_block.py
gheise: removed

mev_inspect/aave_liquidations.py
gheise: Fixed

mev_inspect/aave_liquidations.py
gheise: great tip, done

mev_inspect/aave_liquidations.py
gheise: done

mev_inspect/aave_liquidations.py
gheise: the addresses of the smart contracts through which the liquidation is made

mev_inspect/aave_liquidations.py
gheise: true, but the if is quite lengthy, this seems cleaner imo

mev_inspect/aave_liquidations.py
gheise: PEP 257:
"All modules should normally have docstrings, and all functions and classes exported by a module should also have docstrings."

mev_inspect/aave_liquidations.py
gheise: done

mev_inspect/aave_liquidations.py
gheise: Don't understand the purpose of this one since this is just the execution of the program and some core logic. Passing a function here will make it much more difficult to append to the lists which are all defined locally in the function, so will add more work to pass these as args to a new function

gheise: this was the case with the others but the increment was reasonable; here I feel like that's an unnecessary hurdle and is testable as is

mev_inspect/aave_liquidations.py
lukevs: maybe `LIQUIDATION_CONTRACT_ADDRESSES`?

mev_inspect/aave_liquidations.py
lukevs: it’s still tx

mev_inspect/aave_liquidations.py
lukevs: if you never see that key

like logically in the code there is a case where you never hit that if branch

mev_inspect/aave_liquidations.py
lukevs: it’s not a list of liquidations. transfers to is a list of list. so it’s a list of list of list

mev_inspect/schemas/liquidations.py
gheise: flashloan, dydx, etc

mev_inspect/aave_liquidations.py
lukevs: i think it’s actually less readable because it implies the conditional is more than a boolean somehow

wrapping a bool in an if return True is just reinforcing the existing bool’s value

it’s like me doing absolute_value(1) to make clear 1 is positive

scripts/inspect_block.py
lukevs: it’s still here

mev_inspect/aave_liquidations.py
gheise: I'm not sure how to address this

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,130 @@
...
+
+                # Add the transfer
+                liquidator = next(
+                    (
+                        LIQUIDATION_CONTRACT_ADDRESSES[i]
+                        for i in range(len(LIQUIDATION_CONTRACT_ADDRESSES))
+                        if trace.inputs["recipient"]
+                        == LIQUIDATION_CONTRACT_ADDRESSES[i]
+                    ),
+                    "",
```
lukevs, line 96: Rather than empty string this is better as `None` - then strings can be considered valid and you can explicitly handle the case of it being absent

mev_inspect/aave_liquidations.py
gheise: done

mev_inspect/aave_liquidations.py
lukevs: actually, a clearer way to write this would be:
```
recipient = trace.inputs["recipient"]
if recipient in LIQUIDATION_CONTRACT_ADDRESSES:
    # do whatever
else:
    # do whatever else
```

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,130 @@
...
+LIQUIDATION_CONTRACT_ADDRESSES = [
+    "0x3dfd23A6c5E8BbcFc9581d2E864a68feb6a076d3",
+    "0x7d2768dE32b0b80b7a3454c06BdAc94A69DDc7A9",
+    "0x398eC7346DcD622eDc5ae82352F02bE94C62d119",
+]
+
+
+def find_liquidations_from_traces(
+    traces: List,
+) -> List:
```
lukevs, line 21: sneaky - List of what for both of these

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,130 @@
...
+    """Inspect list of classified traces and identify liquidation"""
+    tx = []
+    liquidations = []
```
lukevs, line 24: type

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,130 @@
...
+    result = []
```
lukevs, line 25: type

mev_inspect/aave_liquidations.py
gheise: I think i's more readable:
is_c(c):
If (c): return True
else: return false

is a very formal way to check if condition c holds, the function follows this. Doesn't seem misleading to me

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,65 @@
...
+        if isinstance(trace, DecodedCallTrace):
+
+            # Check for liquidation and register trace and unique liquidator
+            if (
+                trace.classification == Classification.liquidate
+                and trace.transaction_hash not in seen_transactions
+            ):
+
+                liquidations.append(trace)
+                LIQUIDATION_CONTRACT_ADDRESSES.append(trace.from_address)
```
lukevs, line 38: woah wait so it's not constant?

(shows the value of differentiating with caps for constants - things like this stand out visually)

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,65 @@
...
+                        # This will be the address of the received token
+                    elif input == "_user":
+                        liquidated_user = trace.inputs[input]
+                # Register liquidation
+                result.append(
+                    Liquidation(
+                        liquidated_user=liquidated_user,
+                        collateral_address=collateral_address,
+                        collateral_amount=liquidation_amount,
+                        collateral_source="",
```
lukevs, line 60: bruh

mev_inspect/aave_liquidations.py
lukevs: `liquidated_user` the variable in python won't exist of there is no `_user` value

so when it's used on 57 you'll get a variable doesn't exist exception

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,105 @@
...
+]
+
+
+def is_transfer_to_liquidator(
+    trace: ClassifiedTrace, liquidator: str
+) -> Optional[ClassifiedTrace]:
+    """Check if transfer is to liquidator"""
+
+    if isinstance(trace, DecodedCallTrace):
+        try:
```
lukevs, line 31: try to limit `try` except usage unless truly there's some sort of error (ex: a call to an API times out). better to have explicit control flow

Example:
```
if "recipient" in trace.inputs:
    recipient = trace.inputs["reciplient"]
    if recipient == liquidator and trace.from_address in AAVE_CONTRACT_ADDRESSES:
        return trace

if "dst" in ...
```

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,105 @@
...
+    # AAVE V2
+    "0x7d2768de32b0b80b7a3454c06bdac94a69ddc7a9",
+    # AAVE V1
+    "0x3dfd23a6c5e8bbcfc9581d2e864a68feb6a076d3",
+    # AAVE V2 WETH
+    "0x030ba81f1c18d280636f32af80b9aad02cf0854e",
+]
+
+
+def is_transfer_to_liquidator(
```
lukevs, line 25: `is_x` implies a boolean return

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,105 @@
...
+        if (
+            trace.classification == Classification.liquidate
+            and isinstance(trace, DecodedCallTrace)
+            and trace.transaction_hash not in unique_transaction_hashes
+        ):
+
+            liquidator = trace.from_address
+            unique_transaction_hashes.append(trace.transaction_hash)
+
+            for t in traces:
```
lukevs, line 73: this gets a bruh

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,105 @@
...
+                to_result = is_transfer_to_liquidator(t, liquidator)
```
lukevs, line 74: it looks like here `t` is returned back into `to_result` in which case better to probably have `is_transfer_to_liquidator` actually return a `bool` and just use `t` in the branch rather than introducing an additional variable

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,105 @@
...
+
+    return None
+
+
+def get_liquidations(
+    traces: List[ClassifiedTrace],
+) -> List[Liquidation]:
+
+    """Inspect list of classified traces and identify liquidation"""
+    # liquidation_traces: List[DecodedCallTrace] = []
```
lukevs, line 57: remove?

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,105 @@
...
+    liquidations: List[Liquidation] = []
+    transfers_to: Dict = {}
```
lukevs, line 59: more type

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,105 @@
...
+        try:
+            if (
+                trace.inputs["recipient"] == liquidator
+                and trace.from_address in AAVE_CONTRACT_ADDRESSES
+            ):
+                return trace
+
+        except KeyError:
+            pass
+        try:
```
lukevs, line 40: same here - check in instead of key error try / except

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,105 @@
...
+            unique_transaction_hashes.append(trace.transaction_hash)
+
+            for t in traces:
+                to_result = is_transfer_to_liquidator(t, liquidator)
+                if to_result and not (
+                    to_result.transaction_hash in transfers_to.keys()
+                ):
+                    transfers_to[trace.transaction_hash] = to_result
+
+            try:
```
lukevs, line 80: same here

mev_inspect/schemas/abi.py
```diff
@@ -46,5 +46,19 @@ def get_signature(self) -> str:
         return f"{self.name}({joined_input_types})"
 
 
+class ABIEventDescription(BaseModel):
```
lukevs, line 49: used?

mev_inspect/schemas/liquidations.py
```diff
@@ -0,0 +1,26 @@
...
+
+class Liquidation(BaseModel):
+    liquidated_user: str
+    liquidator_user: str
+    # collateral
+    collateral_token_address: str
+    # collateral_amount: int
+    # debtToCover
+    debt_token_address: str
+    # purchaseAmount
```
lukevs, line 13: i don't think these should be here since they're aave specific

if the field names are unclear such that they need a comment, we should change the names to be clearer

also think this connection is made clear enough in the `get_inspections` itself where you're doing stuff like
```
debt_purchase_amount=trace.inputs["_purchaseAmount"]
```

mev_inspect/schemas/liquidations.py
lukevs: still needs remove

mev_inspect/schemas/liquidations.py
```diff
@@ -0,0 +1,26 @@
+from pydantic import BaseModel
+from mev_inspect.schemas.classified_traces import Protocol
+
+
+class Liquidation(BaseModel):
+    liquidated_user: str
+    liquidator_user: str
+    # collateral
+    collateral_token_address: str
+    # collateral_amount: int
```
lukevs, line 10: missing?

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,105 @@
...
+    liquidations: List[Liquidation] = []
+    transfers_to: Dict = {}
+    unique_transaction_hashes: List = []
+
+    for trace in traces:
+
+        if (
+            trace.classification == Classification.liquidate
+            and isinstance(trace, DecodedCallTrace)
+            and trace.transaction_hash not in unique_transaction_hashes
```
lukevs, line 67: this de-dupe on `transaction_hash` feels like "just gimme whichever trace I see first"

is there a way we can more specifically select the trace we want? how are they different and which one are we actually interested in?

mev_inspect/aave_liquidations.py
lukevs: what are we looking for in this second loop? is there something more explicit we can say here? maybe finding a transfer to the liquidator that's a child trace of this liquidation?

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,105 @@
...
+                    transfers_to[trace.transaction_hash] = to_result
+
+            try:
+                received_amount = int(
+                    transfers_to[trace.transaction_hash].inputs["amount"]
+                )
+            except KeyError:
+                received_amount = int(
+                    transfers_to[trace.transaction_hash].inputs["wad"]
+                )
```
lukevs, line 87: I think this is where you were saying you were getting issues w/ Transfer

Does `Transfer.from_trace(transfers_to[trace.transaction_hash])` give an error?

mev_inspect/aave_liquidations.py
gheise: Got it

mev_inspect/aave_liquidations.py
gheise: I don't understand how? The trace in the branch is the liquidation, t is the associated transfer that is required to derive received_amount. To change traces, the liquidation trace would need to be returned, and it will be missing the subsequent fields

mev_inspect/aave_liquidations.py
gheise: what is more type?

mev_inspect/aave_liquidations.py
lukevs: I'm saying to_result is either `t` or `None` based on the logic of `is_transfer_to_liquidator`

so instead of
```
to_result = is_transfer_to_liquidator(t, liquidator)
if to_result:
    # do stuff with to_result
```

save the extra variable, make `is_transfer_to_liquidator` return a bool, and do
```
if is_transfer_to_liquidator(t, liquidator):
   # do stuff with t
```

mev_inspect/aave_liquidations.py
lukevs: Like what are the types of the Dict
I think this is a `Dict[str, DecodedCallTrace]`

mev_inspect/aave_liquidations.py
gheise: I haven't tried. I was previously doing Transfer.from_trace(trace)

mev_inspect/aave_liquidations.py
gheise: Great feedback thanks will implement

mev_inspect/schemas/liquidations.py
gheise: Got it

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,106 @@
...
+    # AAVE V2
+    "0x7d2768de32b0b80b7a3454c06bdac94a69ddc7a9",
+    # AAVE V1
+    "0x3dfd23a6c5e8bbcfc9581d2e864a68feb6a076d3",
+    # AAVE V2 WETH
+    "0x030ba81f1c18d280636f32af80b9aad02cf0854e",
+]
+
+
+def find_liquidator_payback(
```
lukevs, line 25: find = DB call
get = in memory

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,107 @@
...
+
+                liquidator_payback = find_liquidator_payback(child, liquidator)
+
+                if liquidator_payback:
+
+                    assert isinstance(
+                        liquidator_payback.inputs, Dict
+                    )  # Necessary for mypy to indentify type
+
+                    if "amount" in liquidator_payback.inputs:
```
lukevs, line 82: great place for a function

rather than having multiple branches which assign to `received_amount`

assign to it once
```
received_amount = get_received_amount(...)
```

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,103 @@
...
+            for child in child_traces:
+
+                if child.classification == Classification.liquidate:
+                    traces.remove(child)
+
+                if is_liquidator_payback(child, liquidator):
+
+                    assert isinstance(
+                        child.inputs, Dict
+                    )  # Necessary for mypy to indentify type
```
lukevs, line 76: what type does mypy think this is? my guess is optional[dict]?

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,103 @@
...
+
+                    if "amount" in child.inputs:
```
lukevs, line 78: great chance for a function

each branch of this does an assignment to `received_amount`

can make it more concrete with a single assignment from a function
```
received_amount = get_received_amount(...)
```

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,103 @@
...
+            liquidator = trace.from_address
+
+            child_traces = get_child_traces(
+                trace.transaction_hash, trace.trace_address, traces
+            )
+
+            for child in child_traces:
+
+                if child.classification == Classification.liquidate:
+                    traces.remove(child)
```
lukevs, line 70: doing remove on the list you're iterating over can cause issues

also doing remove with complex objects is a little messy (how does it know which is the object to remove? memory address? comparison for equality of the fields?)

is there another way we can do this?

mev_inspect/aave_liquidations.py
gheise: yes even though it is there and has passed an isinstance

mev_inspect/aave_liquidations.py
gheise: dome

mev_inspect/aave_liquidations.py
lukevs: `trace` passed isinstance, but not child

mev_inspect/aave_liquidations.py
gheise: can drop by index using pop

gheise: or do you mean not remove the child at all

tests/liquidation_test.py
```diff
@@ -1,4 +1,74 @@
...
+from mev_inspect.schemas.liquidations import Liquidation
+from mev_inspect.schemas.classified_traces import Protocol
+from mev_inspect.classifiers.trace import TraceClassifier
+from mev_inspect.block import create_from_block_number
+
+
+class TestAaveLiquidations(unittest.TestCase):
+    def test_single_weth_liquidation(self):
+
+        base_provider = Web3.HTTPProvider("http://162.55.96.141:8546/")
```
lukevs, line 14: we shouldn't be doing any data fetching in tests in case they get run offline / away from where the service can be contacted

check out the arbitrage integration tests for how we leverage saved blocks

bertmiller: Please be careful to not hardcode nodes like this. If possible scrub that node out of the commit history entirely.

tests/liquidation_test.py
```diff
@@ -1,4 +1,74 @@
...
+from web3 import Web3
+
+from mev_inspect.aave_liquidations import get_liquidations
+from mev_inspect.schemas.liquidations import Liquidation
+from mev_inspect.schemas.classified_traces import Protocol
+from mev_inspect.classifiers.trace import TraceClassifier
+from mev_inspect.block import create_from_block_number
+
+
+class TestAaveLiquidations(unittest.TestCase):
```
lukevs, line 11: check out how we use pytest in the arbitrage integration tests instead of unit tests

they're a simpler / don't require a class

tests/liquidation_test.py
lukevs: removed

lukevs (approved):

mev_inspect/aave_liquidations.py
```diff
@@ -0,0 +1,86 @@
...
+                    liquidator_user=liquidator,
+                    debt_purchase_amount=trace.inputs["_purchaseAmount"],
+                    protocol=Protocol.aave,
+                    received_amount=received_amount,
+                    transaction_hash=trace.transaction_hash,
+                    block_number=trace.block_number,
+                )
+            )
+
+    return liquidations
```
bertmiller, line 70: Very clear and readable code

mev_inspect/inspect_block.py
```diff
@@ -75,6 +76,10 @@ def inspect_block(
         delete_arbitrages_for_block(db_session, block_number)
         write_arbitrages(db_session, arbitrages)
 
+    liquidations = get_liquidations(classified_traces)
```
bertmiller, line 79: you're importing "get_liquidations" from "aave_liquidations," which I could see creating confusion in the future

is this for all liquidations? Or just aave liquidations?

if it is for aave liquidations I would rename to something like get_aave_liquidations

bertmiller (commented):
Commented in one place on a function name which gets at the scope of the aave_liquidations module

Is this intended to be more general? Or is it specific to Aave? Depending on the answer to that we should rename Aave or the get_liquidations function, I think

If it is general I'd rename it to be something like `liquidations` and the function `get_liquidations` but if it is Aave specific I'd stick with `aave_liquidations` and change the function to `get_aave_liquidations`

PR #78, opened 2021-09-21
Adjust arbitrage calculation logic

sragss:
Arbitrage logic changes
- Arbitrage logic was [previously dependent](https://github.com/flashbots/mev-inspect-py/blob/main/mev_inspect/arbitrages.py#L89) on pool address chaining. I believe (potentially incorrectly) that this is only the case in uni pools that are swapped via the uni-v2/v3 routers. This approach doesn't work for the case of an arb contract executing the swaps individually against the pools nor other AMMs which do not have the same router paradigm (ex: balancer). 
-[Example tx of scenario described](https://etherscan.io/tx/0xbbb549f93a99a99f4c7aaa6b9c93a1f5f670d13b483be3487b942720c6f08d6e)
- Additionally this adjusts the arb detection logic to sort by transfers upfront (via swap.trace_address)

Testing
- Unit test for ordering by `swap.trace_address`
- Existing arb tests work with new logic
- Spot tested against txs picked up by mev-inspect-rs across balancer and uni in single tx 
- Integration test (`tests/test_arbitrage_integration.py`) was adjusted to a profit amount of .75ETH from 52ETH. I believe this is the correct number unless I'm missing something about how this is calculated ([etherscan of test tx](https://etherscan.io/tx/0x448245bf1a507b73516c4eeee01611927dada6610bf26d403012f2e66800d8f0))

### DIFF
--- a/mev_inspect/arbitrages.py
+++ b/mev_inspect/arbitrages.py
@@ -1,5 +1,6 @@
 from itertools import groupby
-from typing import List, Optional
+from typing import List
+from functools import cmp_to_key
 
 from mev_inspect.schemas.arbitrages import Arbitrage
 from mev_inspect.schemas.swaps import Swap
@@ -23,70 +24,70 @@ def get_arbitrages(swaps: List[Swap]) -> List[Arbitrage]:
 
 
 def _get_arbitrages_from_swaps(swaps: List[Swap]) -> List[Arbitrage]:
-    pool_addresses = {swap.pool_address for swap in swaps}
-
     all_arbitrages = []
 
-    for index, first_swap in enumerate(swaps):
-        other_swaps = swaps[:index] + swaps[index + 1 :]
+    ordered_swaps = _order_swaps_by_trace_order(swaps)
 
-        if first_swap.from_address not in pool_addresses:
-            arbitrage = _get_arbitrage_starting_with_swap(first_swap, other_swaps)
+    grouped_swaps: List[List[Swap]] = []
 
-            if arbitrage is not None:
-                all_arbitrages.append(arbitrage)
+    # An arbitrage is defined as multiple swaps in a row that result in the initial token being returned.
+    # Ex: [WETH -> UNI, UNI -> DAI, DAI -> WETH]
+    # Code below assumes there can be multiple arbitrages per swap set, but they'll be serial.
+    # Ex: [WETH -> UNI, UNI -> WETH, DAI -> SUSHI, SUSHI -> DAI]
+    # Non-working ex: [WETH -> UNI, DAI -> SUSHI, UNI -> WETH, SUSHI -> DAI]
+    for start_swap_index in range(len(ordered_swaps)):
+        current_swap = ordered_swaps[start_swap_index]
+        swap_path: List[Swap] = [current_swap]
+
+        for path_index in range(start_swap_index + 1, len(ordered_swaps)):
+            next_swap = ordered_swaps[path_index]
+            if current_swap.token_out_address == next_swap.token_in_address:
+                swap_path.append(next_swap)
+                current_swap = next_swap
+                if (
+                    swap_path[0].token_in_address == next_swap.token_out_address
+                ):  # Cycle complete
+                    grouped_swaps.append(swap_path)
+                    swap_path = []
+            else:
+                swap_path = []
+
+    for swap_path in grouped_swaps:
+        start_amount = swap_path[0].token_in_amount
+        end_amount = swap_path[-1].token_out_amount
+        profit_amount = end_amount - start_amount
+
+        arb = Arbitrage(
+            swaps=swap_path,
+            block_number=swap_path[0].block_number,
+            transaction_hash=swap_path[0].transaction_hash,
+            account_address=swap_path[0].from_address,
+            profit_token_address=swap_path[0].token_in_address,
+            start_amount=start_amount,
+            end_amount=end_amount,
+            profit_amount=profit_amount,
+        )
+        all_arbitrages.append(arb)
 
     return all_arbitrages
 
 
-def _get_arbitrage_starting_with_swap(
-    start_swap: Swap,
-    other_swaps: List[Swap],
-) -> Optional[Arbitrage]:
-    swap_path = [start_swap]
-    current_swap: Swap = start_swap
-
-    while True:
-        next_swap = _get_swap_from_address(
-            current_swap.to_address,
-            current_swap.token_out_address,
-            other_swaps,
-        )
-
-        if next_swap is None:
-            return None
-
-        swap_path.append(next_swap)
-        current_swap = next_swap
-
-        if (
-            current_swap.to_address == start_swap.from_address
-            and current_swap.token_out_address == start_swap.token_in_address
-        ):
-
-            start_amount = start_swap.token_in_amount
-            end_amount = current_swap.token_out_amount
-            profit_amount = end_amount - start_amount
-
-            return Arbitrage(
-                swaps=swap_path,
-                block_number=start_swap.block_number,
-                transaction_hash=start_swap.transaction_hash,
-                account_address=start_swap.from_address,
-                profit_token_address=start_swap.token_in_address,
-                start_amount=start_amount,
-                end_amount=end_amount,
-                profit_amount=profit_amount,
-            )
-
-    return None
+def _order_swaps_by_trace_order(unordered_swaps: List[Swap]) -> List[Swap]:
+    unordered_swaps.sort(key=cmp_to_key(_compare_trace_address))
+    return unordered_swaps
 
 
-def _get_swap_from_address(
-    address: str, token_address: str, swaps: List[Swap]
-) -> Optional[Swap]:
-    for swap in swaps:
-        if swap.pool_address == address and swap.token_in_address == token_address:
-            return swap
+def _compare_trace_address(a: Swap, b: Swap):
+    if a.trace_address == b.trace_address:
+        return 0
+    for i in range(0, max(len(a.trace_address), len(b.trace_address))):
+        if len(a.trace_address) == i:
+            return -1
+        if len(b.trace_address) == i:
+            return 1
 
-    return None
+        if a.trace_address[i] > b.trace_address[i]:
+            return 1
+        if a.trace_address[i] < b.trace_address[i]:
+            return -1
+    return 0
--- a/tests/test_arbitrage_integration.py
+++ b/tests/test_arbitrage_integration.py
@@ -23,4 +23,4 @@ def test_arbitrage_real_block():
     assert (
         arbitrage.profit_token_address == "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2"
     )
-    assert arbitrage.profit_amount == 53560707941943273628
+    assert arbitrage.profit_amount == 750005273675102326
--- a/tests/test_arbitrages.py
+++ b/tests/test_arbitrages.py
@@ -1,4 +1,7 @@
-from mev_inspect.arbitrages import get_arbitrages
+import itertools
+from typing import List
+
+from mev_inspect.arbitrages import get_arbitrages, _order_swaps_by_trace_order
 from mev_inspect.schemas.swaps import Swap
 from mev_inspect.swaps import (
     UNISWAP_V2_PAIR_ABI_NAME,
@@ -158,3 +161,49 @@ def test_three_pool_arbitrage(get_transaction_hashes, get_addresses):
     assert arbitrage.start_amount == first_token_in_amount
     assert arbitrage.end_amount == first_token_out_amount
     assert arbitrage.profit_amount == first_token_out_amount - first_token_in_amount
+
+
+def test_order_swaps_by_trace_order():
+    trace_address_0 = []
+    trace_address_1 = [0]
+    trace_address_2 = [1]
+    trace_address_3 = [1, 1]
+    trace_address_4 = [2, 1, 3]
+    trace_address_5 = [2, 2]
+
+    permutated_trace_addresses: List[List[int]] = itertools.permutations(
+        [
+            trace_address_0,
+            trace_address_1,
+            trace_address_2,
+            trace_address_3,
+            trace_address_4,
+            trace_address_5,
+        ]
+    )
+
+    for trace_addresses in permutated_trace_addresses:
+        swaps: List[Swap] = []
+        for trace_address in trace_addresses:
+            swap = Swap(
+                abi_name=UNISWAP_V3_POOL_ABI_NAME,
+                transaction_hash="0xfake",
+                block_number=0,
+                trace_address=trace_address,
+                pool_address="0xfake",
+                from_address="0xfake",
+                to_address="0xfake",
+                token_in_address="0xfake",
+                token_in_amount=1,
+                token_out_address="0xfake",
+                token_out_amount=1,
+            )
+            swaps.append(swap)
+        ordered_swaps = _order_swaps_by_trace_order(swaps)
+
+        assert ordered_swaps[0].trace_address == trace_address_0
+        assert ordered_swaps[1].trace_address == trace_address_1
+        assert ordered_swaps[2].trace_address == trace_address_2
+        assert ordered_swaps[3].trace_address == trace_address_3
+        assert ordered_swaps[4].trace_address == trace_address_4
+        assert ordered_swaps[5].trace_address == trace_address_5


mev_inspect/arbitrages.py
```diff
@@ -23,70 +24,70 @@ def get_arbitrages(swaps: List[Swap]) -> List[Arbitrage]:
...
-            other_swaps,
+    ordered_swaps = _order_swaps_by_trace_order(swaps)
+
+    grouped_swaps: List[List[Swap]] = []
+
+    # An arbitrage is defined as multiple swaps in a row that result in the initial token being returned.
+    # Ex: [WETH -> UNI, UNI -> DAI, DAI -> WETH]
+    # Code below assumes there can be multiple arbitrages per swap set, but they'll be serial.
+    # Ex: [WETH -> UNI, UNI -> WETH, DAI -> SUSHI, SUSHI -> DAI]
+    # Non-working ex: [WETH -> UNI, DAI -> SUSHI, UNI -> WETH, SUSHI -> DAI]
```
lukevs, line 37: thoughts on moving to docstring?

tests/test_arbitrage_integration.py
```diff
@@ -23,4 +23,4 @@ def test_arbitrage_real_block():
     assert (
         arbitrage.profit_token_address == "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2"
     )
-    assert arbitrage.profit_amount == 53560707941943273628
```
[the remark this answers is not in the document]
sragss: There's actually 2 arbitrages in this block!

The one I [referenced](https://etherscan.io/tx/0x448245bf1a507b73516c4eeee01611927dada6610bf26d403012f2e66800d8f0). And the [https://etherscan.io/tx/0xfcf4558f6432689ea57737fe63124a5ec39fd6ba6aaf198df13a825dd599bffc]. I have to admit I have no idea what is going on in 0xfcf4...

Here's the full list of swaps
```
[Swap(abi_name='UniswapV3Pool', transaction_hash='0xfcf4558f6432689ea57737fe63124a5ec39fd6ba6aaf198df13a825dd599bffc', block_number=12914944, trace_address=[0, 0], pool_address='0x88e6a0c2ddd26feeb64f039a2c41296fcb3f5640', from_address='0x7ec0b75a98997c927ace6d87958147a105147ea0', to_address='0x0000fee6275dab194ab538a01dd8b18b02b20000', token_in_address='0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48', token_in_amount=283588902010, token_out_address='0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2', token_out_amount=123848351154563483804, protocol=None, error=None), Swap(abi_name='UniswapV3Pool', transaction_hash='0xfcf4558f6432689ea57737fe63124a5ec39fd6ba6aaf198df13a825dd599bffc', block_number=12914944, trace_address=[0, 0, 2, 0], pool_address='0x7ec0b75a98997c927ace6d87958147a105147ea0', from_address='0x99b42f2b49c395d2a77d973f6009abb5d67da343', to_address='0x88e6a0c2ddd26feeb64f039a2c41296fcb3f5640', token_in_address='0x25f8087ead173b73d6e8b84329989a8eea16cf73', token_in_amount=197386080350317054519429, token_out_address='0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48', token_out_amount=283588902010, protocol=None, error=None), Swap(abi_name='UniswapV2Pair', transaction_hash='0xfcf4558f6432689ea57737fe63124a5ec39fd6ba6aaf198df13a825dd599bffc', block_number=12914944, trace_address=[0, 0, 2, 0, 2, 1], pool_address='0x99b42f2b49c395d2a77d973f6009abb5d67da343', from_address='0x0000fee6275dab194ab538a01dd8b18b02b20000', to_address='0x7ec0b75a98997c927ace6d87958147a105147ea0', token_in_address='0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2', token_in_amount=70287643212620210176, token_out_address='0x25f8087ead173b73d6e8b84329989a8eea16cf73', token_out_amount=197386080350317054519429, protocol=None, error=None)]
```

Simplified and ordered txs:
```
- Swap ([0, 0]): 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48 -> 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2
- Swap ([0, 0, 2, 0]): 0x25f8087ead173b73d6e8b84329989a8eea16cf73 -> 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48
- Swap ([0, 0, 2, 0, 2, 1]): 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2 -> 0x25f8087ead173b73d6e8b84329989a8eea16cf73
```

Note that this is reversed based on what we expected from etherscan.
Either trace sort order is wrong (root calls come after child calls) or this contract is doing something very funky. I'll continue to investigate but any thoughts would be awesome!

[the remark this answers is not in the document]
lukevs: Appreciate you digging in and these PRs in general!

For 0xfcf45, I think the trace order is correct and is kind of uncovering an issue with the logic in the comment you left

Arb swaps can happen "out of order" as long as there's a rebalancing of the pool by the end of the transaction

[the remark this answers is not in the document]
sragss: Happy to!

I'm going to rework this to work like the old logic but not be dependent on `swap.pool_address`. May work with a combination of `swap.pool_address` and `swap.from_address`.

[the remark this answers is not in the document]
sragss: I'm seeing two ways to do this
1. Add 2x Transfer objects to each Swap object (as in mev-inspect-rs where a Swap is simply composed of two transfers, resulting in all 4 relevant addresses, rather than 3)
2. Look for cycles of `swap -> ... -> swap` where 
    - `swap.token_in` matches the preceeding `swap.token_out`
    - `swap[0].token_in == swap[-1].token_out`

(1.) is guaranteed to work, but (2.) may be sufficient. The reasoning for why we can't build perfect circuits can be seen in the example txs below. 0x44824 has the to and from addresses as the MEV bot's contract address.

Here's some clearer summaries on the two different txs in this test block that both need to be picked up:
```
ordered swap set for 0x448245bf1a507b73516c4eeee01611927dada6610bf26d403012f2e66800d8f0
Swap: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2 -> 0x25f8087ead173b73d6e8b84329989a8eea16cf73
Pool: 0x99b42f2b49c395d2a77d973f6009abb5d67da343 | from: 0x5e6732d6416d3f3a6574b61ef1c81316cf8d9287 | to: 0x5e6732d6416d3f3a6574b61ef1c81316cf8d9287

Swap: 0x25f8087ead173b73d6e8b84329989a8eea16cf73 -> 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48
Pool: 0xb9c31afc2df2cd49c2a685ff378c21eb7d115541 | from: 0x5e6732d6416d3f3a6574b61ef1c81316cf8d9287 | to: 0x5e6732d6416d3f3a6574b61ef1c81316cf8d9287

Swap: 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48 -> 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2
Pool: 0x88e6a0c2ddd26feeb64f039a2c41296fcb3f5640 | from: 0x5e6732d6416d3f3a6574b61ef1c81316cf8d9287 | to: 0x5e6732d6416d3f3a6574b61ef1c81316cf8d9287
```

```
ordered swap set for 0xfcf4558f6432689ea57737fe63124a5ec39fd6ba6aaf198df13a825dd599bffc
Swap: 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48 -> 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2
Pool: 0x88e6a0c2ddd26feeb64f039a2c41296fcb3f5640 | from: 0x7ec0b75a98997c927ace6d87958147a105147ea0 | to: 0x0000fee6275dab194ab538a01dd8b18b02b20000

Swap: 0x25f8087ead173b73d6e8b84329989a8eea16cf73 -> 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48
Pool: 0x7ec0b75a98997c927ace6d87958147a105147ea0 | from: 0x99b42f2b49c395d2a77d973f6009abb5d67da343 | to: 0x88e6a0c2ddd26feeb64f039a2c41296fcb3f5640

Swap: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2 -> 0x25f8087ead173b73d6e8b84329989a8eea16cf73
Pool: 0x99b42f2b49c395d2a77d973f6009abb5d67da343 | from: 0x0000fee6275dab194ab538a01dd8b18b02b20000 | to: 0x7ec0b75a98997c927ace6d87958147a105147ea0
```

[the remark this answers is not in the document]
lukevs: it looks like in the first one it's printing the same address for to and from

what's the difference between options 1 and 2 in practice? I'd vote 2 since it stays derivative of the swap

the from_address and to_address for the swap should be the address putting tokens in and the one where the tokens are coming out, so it should be the same as looking at the transfer level. Right now we're indifferent whether those addresses are a contract address or an EOA (better to be able to see where people are keeping tokens)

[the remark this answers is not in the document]
sragss: Ok, I've reworked this substantially to implement option 2. Updating the PR description to explain the new algorithm and current short-falls.

Basically we have two types of different arbitrages we need to deal with

Uni router style: MEV_BOT -> Pool A -> Pool B -> MEV_BOT
Standard style: MEV_BOT -> Pool A -> MEV_BOT -> Pool B -> MEV_BOT

The new detection strategy doesn't depend entirely depend on these transfer addresses so it handles both. 

(I'll try to outline this thoroughly in the PR description)

mev_inspect/arbitrages.py
```diff
@@ -23,70 +24,70 @@ def get_arbitrages(swaps: List[Swap]) -> List[Arbitrage]:
...
-                account_address=start_swap.from_address,
-                profit_token_address=start_swap.token_in_address,
-                start_amount=start_amount,
-                end_amount=end_amount,
-                profit_amount=profit_amount,
-            )
 
-    return None
+def _order_swaps_by_trace_order(unordered_swaps: List[Swap]) -> List[Swap]:
+    unordered_swaps.sort(key=cmp_to_key(_compare_trace_address))
```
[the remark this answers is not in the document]
sragss: Lol good call. Didn't even think of trying that. No longer using this code either way.

mev_inspect/arbitrages.py
sragss: Done.

tests/test_arbitrage_integration.py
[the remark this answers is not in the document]
sragss: Details in PR description!

mev_inspect/arbitrages.py
```diff
@@ -23,70 +23,124 @@ def get_arbitrages(swaps: List[Swap]) -> List[Arbitrage]:
...
+        for potential_end_swap in swaps:
+
+            # Checking by profitability can over-discover due to MEV-bot dust. Ex: 3 pool arb, second tx does not use all of output of first, we'd classify two start/end points. These have to be filtered separately.
+            if (
+                potential_start_swap.token_in_address
+                == potential_end_swap.token_out_address
+                and potential_start_swap.token_in_amount
+                < potential_end_swap.token_out_amount
+            ):
+                results.append((potential_start_swap, potential_end_swap))
```
lukevs, line 94: is there a way we could keep non-profitable arbs? - it can be neat to see that people that ended up executing something that net lost them money

for the case where someone doesn't use all of the output in an intermediate step, why does that result in two start / ends?

mev_inspect/arbitrages.py
```diff
@@ -23,70 +23,124 @@ def get_arbitrages(swaps: List[Swap]) -> List[Arbitrage]:
...
-            if arbitrage is not None:
-                all_arbitrages.append(arbitrage)
+    all_arbitrages = []
 
-    return all_arbitrages
+    start_ends = _get_all_start_end_swaps(swaps)
+    if len(start_ends) == 0:
+        return []
 
+    filtered_start_ends = filter(
```
lukevs, line 42: what does this step get us?

I think as long as all of the swaps happened in the same transaction and form a circuit, it shouldn't matter the ordering within the trace

mev_inspect/arbitrages.py
```diff
@@ -23,70 +23,124 @@ def get_arbitrages(swaps: List[Swap]) -> List[Arbitrage]:
...
+        lambda start_end: _swaps_have_valid_order_wrt_trace_address(list(start_end)),
+        start_ends,
+    )
+    for start_end in filtered_start_ends:
+        start = start_end[0]
+        end = start_end[1]
```
lukevs, line 48: small - could do
```
for [start, end] in filtered_start_ends:
    ...
```

mev_inspect/arbitrages.py
lukevs: Just saw the piece in the PR mentioning it

I would think these
```
one of the intermediate transactions in the swap route does not use all of the output of the previous swap
one of the intermediate transactions in the swap route uses more than the output of the previous swap
```
are still arbs - we might need some better ways to describe what's happening though

mev_inspect/arbitrages.py
```diff
@@ -23,70 +23,124 @@ def get_arbitrages(swaps: List[Swap]) -> List[Arbitrage]:
...
 
-def _get_arbitrage_starting_with_swap(
-    start_swap: Swap,
-    other_swaps: List[Swap],
-) -> Optional[Arbitrage]:
-    swap_path = [start_swap]
-    current_swap: Swap = start_swap
+        remaining_swaps = [swap for swap in swaps if swap != start and swap != end]
```
lukevs, line 50: I think it won't end up mattering here, but are these equality checks by attribute value comparisons or by the underlying object id?

it might be ideal to just return these as part of the generation of the start end pairs similar to the original version `(start_swap, end_swap, other_swaps)`

mev_inspect/arbitrages.py
```diff
@@ -23,70 +23,124 @@ def get_arbitrages(swaps: List[Swap]) -> List[Arbitrage]:
...
+    # If the path is complete, return
+    if start_swap.token_out_address == end_swap.token_in_address:
+        return [[start_swap, end_swap]]
+    elif len(other_swaps) == 0:
+        return []
+
+    # Collect all potential next steps, check if valid, recursively find routes from next_step to end_swap
+    routes: List[List[Swap]] = []
+    for i in range(len(other_swaps)):
+        potential_next_swap = other_swaps[i]
```
lukevs, line 113: looks like the i is unused:
```
for potential_next_swap in other_swaps:
   ...
```

mev_inspect/arbitrages.py
sragss: Notes on why relying on profitability is necessary in comment#<FILL THIS IN SAM>.

On not using all the output-
If you have `A->B`/`B->C`/`C->A` and you're trying to find a start/end based on profitability and you have
- `(A->B).a_in < (C->A).a_out` - due to successful arbitrage
- `(B->C).b_in < (A->B).b_out` - due to dust / rounding error
There is no way to tell which is the valid start and end. Of course you could add a config that says drop arbs below a certain amount of profitability, but that's its own can of worms.

mev_inspect/arbitrages.py
sragss: Yeah so currently we still pick these up. Vast majority of the time this happens is due to one of the following two cases:
- Rounding error in the contract: Sometimes MEV bot contracts only care about the first 12 decimals for example.
- Contract is doing A -> B -> A, happens to have some extra B sitting around when it does the swap and just takes all of its B and puts it in the B->A swap.
From my testing it's usually an insignificant sum (10^-6 in size), so I don't think it's worth worrying about unless we find some seriously mis-evaluated arbs down the road.

## Why the filtering start_ends matters:
If we have A -> B -> C -> A, this comes out to these swaps:
1. A->B
2. B->C
3. C->A

Without any assumptions this can be routed as:
`A->B->C->A`, `B->C->A->B`, `C->A->B->C`
So we need to add some assumptions to find a way to choose the starting and ending point. 

**We can't use `to` and `from` addresses of the swaps** because of the case where the bot contract is executing swaps individually:
- Case where this would work: `BOT->Pool A/B->Pool B/C->Pool A/C->BOT`
- Case where this would break: `BOT->Pool A/B->BOT` / `BOT->Pool B/C->BOT` / `BOT->Pool A/C->BOT`

The next obvious way would be based on transaction ordering. Unfortunately **we can't exclusively rely on this due to flash swaps** (previously referenced [tx](https://etherscan.io/tx/0xfcf4558f6432689ea57737fe63124a5ec39fd6ba6aaf198df13a825dd599bffc), uni flash swap [docs](https://docs.uniswap.org/protocol/V2/concepts/core-concepts/flash-swaps)) which allow the actual swap events and transfer events to be out of order. 

Thus we rely on profitability for detecting the start/end swap. Two notes here
- `mev-inspect-rs` does this as well
- From my experience (plz correct me if i'm wrong) no successful MEV bot contract is dumb enough to execute a series of swaps (and pay the miner) if it is not sum profitable after on-chain execution.

In summary:
- Relying on transaction ordering via `trace_address` isn't perfect
- Relying on profitability of start vs final swap isn't perfect

The current solution is to semi rely on both and it seems to work quite well in practice.
- Rely on profitability for start/end swaps
- Rely on transaction ordering to the extent that swaps at least happen at the same trace level. This allows 
  - (most) out of order flash swaps through the filter 
  - filters out the arbs where an intermediate swap is detected as a start/end swap due to `prev_swap.token_out_amount != next_swap.token_in_amount` from dust discrepancies (details in COMMENT#FIL THIS IN SAM

mev_inspect/arbitrages.py
sragss: Nice– done.

mev_inspect/arbitrages.py
sragss: Good call–Adjusted to "is not" to compare by identity. Just have to be careful not to copy the swap objects objects in later updates to the logic.

I also changed the naming of that list to `potential_intermediate_swaps`. I think it's more readable to filter for the `potential_intermediate_swaps` afterwards rather than expanding the tuple and returning the additional set.

mev_inspect/arbitrages.py
sragss: Done.

mev_inspect/arbitrages.py
lukevs: re:
> From my experience (plz correct me if i'm wrong) no successful MEV bot contract is dumb enough to execute a series of swaps (and pay the miner) if it is not sum profitable after on-chain execution.

some bots exclude the check for gas savings...

lukevs: Instead of using profitability or trace ordering to determine where to start, is it possible to just assume the start / end will be whichever address isn't a pool? I think that would work for both cases

mev_inspect/arbitrages.py
lukevs: Yeah agreed on being "careful not to copy" - that concern is why I suggest generating the list of other swaps alongside since then it can be done using indices instead of a filtering of the original list like here https://github.com/flashbots/mev-inspect-py/blob/main/mev_inspect/arbitrages.py#L31

mev_inspect/arbitrages.py
```diff
@@ -23,70 +23,121 @@ def get_arbitrages(swaps: List[Swap]) -> List[Arbitrage]:
...
+                for next_swap_route in next_swap_routes:
+                    next_swap_route.insert(0, start_swap)
+                    routes.append(next_swap_route)
+    return routes
+
+
+def _swaps_have_valid_order_wrt_trace_address(swaps: List[Swap]) -> bool:
+    """
+    Checks if a list of swaps have a potentially valid order exclusively based on `swap.trace_address`.
+    """
```
lukevs, line 128: If I'm reading right, valid means that the order in `swaps` is the same order as they occurred in the trace - is that right?

If so, because trace ordering is the natural ordering when comparing lists of ints, you could do something like this which I think more reads like that description:
```
swaps_trace_addresses = [swap.trace_address for swap in swaps]
ordered_trace_addresses = list(sorted(swaps_trace_addresses))

return swaps_trace_addresses == ordered_trace_addresses
```

sragss: This is slightly less specific to allows for **some** out of order txs (this is from `test_arbitrages.py`). Basically it still allows for flash swaps. 

```
    # Block 12914944
    swaps = [
        create_generic_swap(trace_address=[0, 0, 2, 0, 2, 1]),
        create_generic_swap(trace_address=[0, 0, 2, 0]),
        create_generic_swap(trace_address=[0, 0]),
    ]
```

Regardless, I put a new proposal above that won't rely on this kind of logic.

mev_inspect/arbitrages.py
sragss: Ok, fair enough, we can try to include non-profitable. Here's an [ex tx](https://etherscan.io/tx/0x4f2ec102adc25ceb8440a558e3d82e97a4be3ba839dd055db2749ba194d44d8b). 

99+% of arbs I've looked at look like the following two cases:
Case 1 (fully routed. one transfer from the bot, one transfer to the bot):
`Bot -> A/B -> B/C -> C/A -> Bot`

Case 2 (Bot is always intermediary):
`Bot -> A/B -> Bot -> B/C -> Bot -> A/C -> Bot`

Case 1:
- These are easy, we can only calculate one route to and from the bot, making starting ending swap obvious
- (These are also commonly out of order based on trace_address due to flash swaps: [ex](https://etherscan.io/tx/0x0542cbf5ab7ed4b616d01de4bfa4f870a4fc7b1ed7288322fe9cddf0dbd5149b))

Case 2:
- These are the issue, for 3 swaps, we can calculate 3 different starting/ending points and therefore 3 routes
- (In the hundreds of examples I've checked, these are in order: [ex](https://etherscan.io/tx/0x4f2ec102adc25ceb8440a558e3d82e97a4be3ba839dd055db2749ba194d44d8b), but I can imagine a case where they are not)

Proposed solution: Try to fully route a set of swaps into arbs (not relying on profitability). If we only get one valid route, we assume a valid arb (this would cover all of Case 1). If we get multiple valid routes (Case 2) with overlapping swaps, we assume only one is valid and make that decision based on trace_address ordering. Alternatively we could make that decision based on profitability. Let me know what you think!

lukevs: Again I really appreciate you writing these up 🙏 This is super helpful

For case 2, I think the resulting arbitrage object will look the same regardless of which we choose as the start right? As long as we pick a point which isn't a pool address. So first in transaction should be solid.

lukevs: Overall this sounds solid though and nice and simple!

sragss: Yes no worries!

>For case 2, I think the resulting arbitrage object will look the same regardless of which we choose as the start right?

No because the `profit_token` `profit_amount` and `swaps` order will be different. 

I'll rewrite to the new algorithm and push. Will be up shortly.

sragss: Updated. Also updated the PR description.

PR #55, opened 2021-08-24
Compound & forks liquidations

taarushv:

PR #83, opened 2021-09-29
Compound + Forks

taarushv:
Supersedes PR #55 because that one was pre-kube setup


mev_inspect/compound_liquidations.py
```diff
@@ -0,0 +1,113 @@
...
+#             classified_trace.to_address == liquidation_contract
+#             and classified_trace.function_name == "liquidateBorrow"
+#         ):
+#             ## check if tx originates from searcher eoa or contract
+#             if tx.to_address.lower() == classified_trace.from_address:
+#                 source = LiquidationCollateralSource.searcher_contract
+#             elif tx.from_address.lower() == classified_trace.from_address:
+#                 source = LiquidationCollateralSource.searcher_eoa
+#             ## flashloan providers identified here
+#     return source
```
bertmiller, line 61: What's with the commented out code

mev_inspect/compound_liquidations.py
```diff
@@ -0,0 +1,113 @@
...
+rpc = os.getenv("RPC_URL")
+
+if rpc is None:
+    raise RuntimeError("Missing environment variable RPC_URL")
+else:
+    w3 = Web3(Web3.HTTPProvider(rpc))
+
+V2_COMPTROLLER_ADDRESS = "0x3d9819210A31b4961b30EF54bE2aeD79B9c9Cd3B"
+V2_C_ETHER = "0x4Ddc2D193948926D02f9B1fE9e1daa0718270ED5"
+WETH_ADDRESS = "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2"
```
bertmiller, line 19: small thing but do we have WETH somewhere else?

if we go to multiple EVM chains at some point it'd be really helpful to be able to change WETH's address in 1 place

mev_inspect/historical_price.py
```diff
@@ -0,0 +1,182 @@
+import os
+
+from mev_inspect.schemas.classified_traces import Protocol
+from mev_inspect.abi import get_raw_abi
+from web3 import Web3
+
+WETH_ADDRESS = "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2"
+USDC_ADDRESS = "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48"
```
bertmiller, line 8: same comment as before re: weth/usdc

mev_inspect/historical_price.py
```diff
@@ -0,0 +1,182 @@
...
+ZERO_ADDRESS = "0x0000000000000000000000000000000000000000"
+
+UNISWAP_ROUTER = "0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D"
+UNISWAP_FACTORY = "0x5C69bEe701ef814a2B6a3EDD4B1652CB9cc5aA6f"
+SUSHISWAP_ROUTER = "0xd9e1cE17f2641f24aE83637ab66a2cca9C378B9F"
+SUSHISWAP_FACTORY = "0xC0AEe478e3658e2610c5F7A4A2E1777cE9e4f2Ac"
```
bertmiller, line 14: probably set somewhere in the Uni classifiers, no?

tests/test_liquidations.py
```diff
@@ -0,0 +1,35 @@
...
+#     #         transaction_hash=transaction_hash,
+#     #         block_number=block_number,
+#     #     )
+#     # ]
+
+#     # block = load_test_block(block_number)
+#     # trace_classifier = TraceClassifier()
+#     # classified_traces = trace_classifier.classify(block.traces)
+#     # result = get_cether_liquidations(classified_traces)
+#     # assert(result == liquidations)
```
bertmiller, line 35: what's going on here?

bertmiller (commented):
Few comments, not sure if this was in a place where you wanted comments or not

tests/test_liquidations.py
taarushv: had to comment out until I got cache files for a few blocks of out the kube setup (as they'd fail), fixed that, will push new commits + rebase off main to use the common liquidation types

mev_inspect/compound_liquidations.py
lukevs: it's in `mev_inspect/classifiers/specs/weth.py`

could change it to a constant there and import

mev_inspect/compound_liquidations.py
```diff
@@ -0,0 +1,113 @@
...
+        Web3.toChecksumAddress(underlying_token_address)
+    )
+    decimals_in_ctoken = get_erc20_token_decimals(c_token_address)
+    return raw_exchange_rate / (
+        10 ** (18 + decimals_in_underlying - decimals_in_ctoken)
+    )
+
+
+# CEther liquidations for now, CToken liquidations have a slightly different format
+def get_cether_liquidations(
```
lukevs, line 85: Can you add to `inspect_block`?

ex: https://github.com/flashbots/mev-inspect-py/blob/main/mev_inspect/inspect_block.py#L79

mev_inspect/compound_liquidations.py
```diff
@@ -0,0 +1,113 @@
...
+from mev_inspect.historical_price import get_erc20_token_decimals
+from mev_inspect.abi import get_raw_abi
+import os
+
+rpc = os.getenv("RPC_URL")
+
+if rpc is None:
+    raise RuntimeError("Missing environment variable RPC_URL")
+else:
+    w3 = Web3(Web3.HTTPProvider(rpc))
```
lukevs, line 15: instead of instantiating `w3`'s globally, there's one created once in `inspect_block` that should be passed down through `get_cether_liquidations`

mev_inspect/compound_liquidations.py
```diff
@@ -0,0 +1,113 @@
...
+
+V2_COMPTROLLER_ADDRESS = "0x3d9819210A31b4961b30EF54bE2aeD79B9c9Cd3B"
+V2_C_ETHER = "0x4Ddc2D193948926D02f9B1fE9e1daa0718270ED5"
+WETH_ADDRESS = "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2"
+
+# cToken=>Token mapping (cDAI=>DAI)
+# useful for finding out the underlying asset address of the cToken seized
+def get_all_comp_markets():
```
lukevs, line 23: How does this work for backfills? Do they ever get changed between blocks?

Also can you add a return type

mev_inspect/compound_liquidations.py
```diff
@@ -0,0 +1,113 @@
...
+    return raw_exchange_rate / (
+        10 ** (18 + decimals_in_underlying - decimals_in_ctoken)
+    )
+
+
+# CEther liquidations for now, CToken liquidations have a slightly different format
+def get_cether_liquidations(
+    classified_traces: list[ClassifiedTrace],
+) -> List[Liquidation]:
+    comp_markets = get_all_comp_markets()
```
lukevs, line 88: can this function take in `comp_markets` instead?

that would let this be fully static instead of making an internal network call which makes scaling easier

mev_inspect/compound_liquidations.py
```diff
@@ -0,0 +1,113 @@
...
+    c_token_address: str, block_number: int
+) -> float:
+    comp_v2_ctoken_abi = get_raw_abi("CToken", Protocol.compound_v2)
+    ctoken_instance = w3.eth.contract(address=c_token_address, abi=comp_v2_ctoken_abi)
+    raw_exchange_rate = ctoken_instance.functions.exchangeRateCurrent().call(
+        block_identifier=block_number
+    )
+    # format based on decimals in ctoken and the underlying token
+    # see "Interpreting Exchange Rates" https://compound.finance/docs#protocol-math
+    underlying_token_address = get_all_comp_markets()[c_token_address.lower()]
```
lukevs, line 74: similar comment here on passing markets instead of fetching to separate network calls from the logic operating on the data

mev_inspect/abi.py
```diff
@@ -11,6 +11,21 @@
 THIS_FILE_DIRECTORY = Path(__file__).parents[0]
 ABI_DIRECTORY_PATH = THIS_FILE_DIRECTORY / "abis"
 
+# raw abi, for instantiating contract for queries (as opposed to classification)
+def get_raw_abi(abi_name: str, protocol: Optional[Protocol]) -> str:
```
lukevs, line 15: does `get_abi(...).json()` work?

mev_inspect/abi.py
```diff
@@ -11,6 +11,21 @@
...
+        ABI_DIRECTORY_PATH / abi_filename
+        if protocol is None
+        else ABI_DIRECTORY_PATH / protocol.value / abi_filename
+    )
+
+    if abi_path.is_file():
+        with abi_path.open() as abi_file:
+            return abi_file.read()
+
+    return ""
```
lukevs, line 27: better to return `None` here. Then we know if it's a string it's an ABI

mev_inspect/historical_price.py
```diff
@@ -0,0 +1,182 @@
...
+    factory_instance = w3.eth.contract(address=UNISWAP_FACTORY, abi=factory_abi)
+    pair_address = factory_instance.functions.getPair(token_0, token_1).call()
+    # 0x0 is returned when the pair does not have a pool
+    if pair_address != ZERO_ADDRESS:
+        return pair_address
+    else:
+        return None
+
+
+def get_sushiswap_pair_pool(token_0, token_1):
```
lukevs, line 46: return types for these

taarushv:
Tested that inspector/db inserts work using the common structure: 

```
fireball@Fireballs-MacBook-Pro mev-inspect-py % kubectl exec deploy/mev-inspect-deployment -- poetry run inspect-block 13326607                
Skipping virtualenv creation, as specified in config file.
Cache for block 13326607 did not exist, getting data
INFO:mev_inspect.inspect_block:Total traces: 1350
INFO:mev_inspect.inspect_block:Total transactions: 214
INFO:mev_inspect.inspect_block:Returned 1349 classified traces
INFO:mev_inspect.inspect_block:Found 11 swaps
INFO:mev_inspect.inspect_block:Found 1 arbitrages
INFO:mev_inspect.inspect_block:Found 0 AAVE liquidations
INFO:mev_inspect.inspect_block:Found 1 COMP liquidations
```

mev_inspect/abi.py
```diff
@@ -11,6 +11,21 @@
...
+        ABI_DIRECTORY_PATH / abi_filename
+        if protocol is None
+        else ABI_DIRECTORY_PATH / protocol.value / abi_filename
+    )
+
+    if abi_path.is_file():
+        with abi_path.open() as abi_file:
+            return abi_file.read()
+
+    return None
```
lukevs, line 27: I think this got lost in the changes, but does `get_abi(...).json()` work instead?

If not, it'd be better for the two to share underlying logic rather than duplicating.

ex:
```
def get_abi(abi_name: str, protocol: Optional[Protocol]) -> Optional[ABI]:
    abi_json = get_raw_abi(abi_name, protocol)
    if abi_json is None:
        return None
    else:
        return parse_obj_as(ABI, abi_json)
```

taarushv: @lukevs yeah, the .json() didn't work but I can remove the duplication with shared functions

tests/test_compound.py
```diff
@@ -0,0 +1,116 @@
...
+            protocol=Protocol.compound_v2,
+            transaction_hash=transaction_hash,
+            trace_address=[1],
+            block_number=block_number,
+        )
+    ]
+    block = load_test_block(block_number)
+    trace_classifier = TraceClassifier()
+    classified_traces = trace_classifier.classify(block.traces)
+    comp_markets = get_all_comp_markets(w3)
```
lukevs, line 40: similar to saving the block data, is it possible to save this data to make tests fully static?

otherwise RPC / HTTP issues could cause these tests to fail when the code itself is working fine

also prevents us from getting that sweet sweet robert airplane code ✈️

mev_inspect/compound_liquidations.py
```diff
@@ -0,0 +1,126 @@
...
+from mev_inspect.schemas.liquidations import Liquidation
+from mev_inspect.classifiers.specs import WETH_ADDRESS
+from mev_inspect.abi import get_raw_abi
+from web3 import Web3
+
+V2_COMPTROLLER_ADDRESS = "0x3d9819210A31b4961b30EF54bE2aeD79B9c9Cd3B"
+V2_C_ETHER = "0x4Ddc2D193948926D02f9B1fE9e1daa0718270ED5"
+
+# helper, only queried once in the beginning (inspect_block)
+def get_all_comp_markets(w3: Web3) -> Dict[str,str]:
```
lukevs, line 24: a naming convention I've seen and like is:
- `get_` for in memory (ex: `def get_name(): return self._name`) - most fast
- `find_` for database queries - medium speed
- `fetch_` for network calls - most slow

That helps to see that something like:
```
for i in range(10000):
    ... = get_name()
```
is fine, but that

```
for i in range(10000):
    ... = fetch_name()
```
would be a bottleneck

for that naming then this would be `fetch_all_comp_markets`

mev_inspect/compound_liquidations.py
```diff
@@ -0,0 +1,126 @@
...
+    c_token_mapping = {}
+    comp_v2_comptroller_abi = get_raw_abi("Comptroller", Protocol.compound_v2)
+    comptroller_instance = w3.eth.contract(
+        address=V2_COMPTROLLER_ADDRESS, abi=comp_v2_comptroller_abi
+    )
+    markets = comptroller_instance.functions.getAllMarkets().call()
+    for c_token in markets:
+        # make an exception for cETH (as it has no .underlying())
+        if c_token != V2_C_ETHER:
+            comp_v2_ctoken_abi = get_raw_abi("CToken", Protocol.compound_v2)
```
lukevs, line 34: load the ABI itself once instead of on each iteration

mev_inspect/compound_liquidations.py
```diff
@@ -0,0 +1,126 @@
...
+            if trace.abi_name =='CEther':
+                liquidations.append(
+                    Liquidation(
+                        liquidated_user=trace.inputs["borrower"],
+                        collateral_token_address=WETH_ADDRESS, # WETH since all cEther liquidations provide Ether
+                        debt_token_address=comp_markets[c_token_collateral],
+                        liquidator_user=seize_trace.inputs["liquidator"],
+                        debt_purchase_amount=trace.value,
+                        protocol=Protocol.compound_v2,
+                        received_amount=seize_trace.inputs["seizeTokens"] * underlying_exchange_rate,
```
lukevs, line 97: the intention for `received_amount` was to be denominated in the received token

I know for aave @gheise is working on just looking at the underlying transfer traces to get this value

if that's possible here it would save us from having to look up exchange rates

mev_inspect/compound_liquidations.py
```diff
@@ -0,0 +1,126 @@
...
+                        transaction_hash=trace.transaction_hash,
+                        trace_address=trace.trace_address,
+                        block_number=trace.block_number,
+                    )
+                )
+    return liquidations
+
+
+def _get_seize_call(child_traces: List[ClassifiedTrace]) -> Optional[ClassifiedTrace]:
```
lukevs, line 121: small but better to call this `traces` if the fact that they're children doesn't matter for the job the function does

mev_inspect/inspect_block.py
```diff
@@ -82,12 +85,19 @@ def inspect_block(
...
-    liquidations = get_aave_liquidations(classified_traces)
-    logger.info(f"Found {len(liquidations)} liquidations")
+    aave_liquidations = get_aave_liquidations(classified_traces)
+    logger.info(f"Found {len(aave_liquidations)} AAVE liquidations")
+
+    comp_markets = get_all_comp_markets(w3)
+    compound_liquidations = get_compound_liquidations(
+        classified_traces, comp_markets, w3
+    )
+    logger.info(f"Found {len(compound_liquidations)} COMP liquidations")
```
lukevs, line 95: can we combine these as a single beautiful `get_liquidations` interface?

mev_inspect/compound_liquidations.py
```diff
@@ -0,0 +1,126 @@
...
+        if c_token != V2_C_ETHER:
+            comp_v2_ctoken_abi = get_raw_abi("CToken", Protocol.compound_v2)
+            ctoken_instance = w3.eth.contract(address=c_token, abi=comp_v2_ctoken_abi)
+            underlying_token = ctoken_instance.functions.underlying().call()
+            c_token_mapping[
+                c_token.lower()
+            ] = underlying_token.lower()  # make k:v lowercase for consistancy
+    return c_token_mapping
+
+def get_underlying_ctoken_exchange_rate(
```
lukevs, line 42: get => fetch - from other comment

mev_inspect/compound_liquidations.py
```diff
@@ -0,0 +1,126 @@
...
+    decimals_in_underlying = get_erc20_token_decimals(
+        underlying_token_address,
+        w3
+    )
+    decimals_in_ctoken = get_erc20_token_decimals(c_token_address, w3)
+    return raw_exchange_rate / (
+        10 ** (18 + decimals_in_underlying - decimals_in_ctoken)
+    )
+
+def get_erc20_token_decimals(token_address: str, w3: Web3) -> int:
```
lukevs, line 62: get => fetch

lukevs:
Mind running pre-commit when you have a chance? I think there might be some type errors that need to be resolved that mypy should catch

mev_inspect/compound_liquidations.py
```diff
@@ -0,0 +1,130 @@
...
+        if c_token != V2_C_ETHER:
+            ctoken_instance = w3.eth.contract(address=c_token, abi=comp_v2_ctoken_abi)
+            underlying_token = ctoken_instance.functions.underlying().call()
+            c_token_mapping[
+                c_token.lower()
+            ] = underlying_token.lower()  # make k:v lowercase for consistancy
+    return c_token_mapping
+
+
+def fetch_underlying_ctoken_exchange_rate(
```
lukevs, line 38: It looks like this is unused

PR #87, opened 2021-10-04
Historical price for tokens (in ETH/USDC)

taarushv:
PR split from #83 to its own


lukevs:
This is now handled by prices

PR #89, opened 2021-10-05
Add received_token_address to liquidations 

gheise:
Add the address of the received token to liquidation object. This will facilitate linking swaps and arbitrages with liquidations through token address, and the method used will remove the need to create the lookup of protocol tokens by using the liquidator payback to determine the token arriving at the liquidator.


mev_inspect/aave_liquidations.py
```diff
@@ -51,7 +51,17 @@ def get_aave_liquidations(
...
+
+            if payback_transfer:
+                assert isinstance(payback_transfer, ERC20Transfer)
+                received_amount = payback_transfer.amount
+                received_token_address = payback_transfer.token_address
+            else:
+                received_amount = 0
+                received_token_address = trace.inputs["_collateral"]
+
+            print(received_token_address)
```
lukevs, line 64: should these be here?

mev_inspect/aave_liquidations.py
```diff
@@ -62,17 +72,21 @@ def get_aave_liquidations(
                     debt_purchase_amount=trace.inputs["_purchaseAmount"],
                     protocol=Protocol.aave,
                     received_amount=received_amount,
+                    received_token_address=received_token_address,
                     transaction_hash=trace.transaction_hash,
                     trace_address=trace.trace_address,
                     block_number=trace.block_number,
                 )
             )
+    print(liquidations)
```
lukevs, line 81: print

mev_inspect/aave_liquidations.py
```diff
@@ -51,7 +51,17 @@ def get_aave_liquidations(
                 trace.transaction_hash, trace.trace_address, traces
             )
 
-            received_amount = _get_liquidator_payback(child_traces, liquidator)
+            payback_transfer = _get_liquidator_payback(child_traces, liquidator)
+
+            if payback_transfer:
+                assert isinstance(payback_transfer, ERC20Transfer)
```
lukevs, line 57: is this necessary? i'd think it'd fail down in the function

`ERC20Transfer.from_trace` inside the function should definitely return `ERC20Transfer`

mev_inspect/aave_liquidations.py
```diff
@@ -51,7 +51,17 @@ def get_aave_liquidations(
...
+                received_amount = payback_transfer.amount
+                received_token_address = payback_transfer.token_address
```
lukevs, line 59: your favorite :) 

if someone changes one of these branches to not set `received_token_address` then it becomes an awkward variable does not exist down below

mev_inspect/aave_liquidations.py
```diff
@@ -51,7 +51,17 @@ def get_aave_liquidations(
                 trace.transaction_hash, trace.trace_address, traces
             )
 
-            received_amount = _get_liquidator_payback(child_traces, liquidator)
+            payback_transfer = _get_liquidator_payback(child_traces, liquidator)
+
+            if payback_transfer:
```
lukevs, line 56: `if payback_transfer is not None` - explicit in your expectations

one time in a project I had an error that came down to an `if x` and it turned out that we thought x was `None` but it was somehow False (the types were crazy in this repo)

mev_inspect/models/liquidations.py
```diff
@@ -12,6 +12,7 @@ class LiquidationModel(Base):
     debt_token_address = Column(String, nullable=False)
     debt_purchase_amount = Column(Numeric, nullable=False)
     received_amount = Column(Numeric, nullable=False)
+    received_token_address = Column(String, nullable=False)
```
lukevs, line 15: is it intentional that this is nullable=False but the DB is nullable=True?

If it was intentional, wanted to callout that it's smart and you're crushing (keep the code level non-null)

If it was an accident, want to call out to say keep doing it :)

alembic/versions/205ce02374b3_add_received_collateral_address_to_liquidations.py
```diff
@@ -0,0 +1,27 @@
...
+revision = "205ce02374b3"
+down_revision = "c8363617aa07"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
+        "liquidations",
+        sa.Column("received_token_address", sa.String(256), nullable=True),
```
lukevs, line 22: what happens if i call inspect block after merging the code and before applying the migration?

since the field is in the model, i'm wondering if it'll say `received_token_address` doesn't exist on save or if it'll just ignore the field on the model

mev_inspect/aave_liquidations.py
gheise: Yes, mypy

mev_inspect/aave_liquidations.py
gheise: True, done

mev_inspect/models/liquidations.py
gheise: Was intentional for sure, seemed like an important feature

alembic/versions/205ce02374b3_add_received_collateral_address_to_liquidations.py
lukevs: Have you tried it? I think SQLAlchemy will attempt to write to it

Can test by downgrading to remove the column

alembic/versions/205ce02374b3_add_received_collateral_address_to_liquidations.py
```diff
@@ -0,0 +1,27 @@
...
+    )
+
+
+def downgrade():
+    op.drop_table("liquidations")
```
lukevs, line 27: Just noticed this - should be a remove column, not dropping the table

Always good to test both downgrade and upgrade by upgrading, downgrading, then upgrading again
```
alembic upgrade head
alembic downgrade -1
alembic upgrade head
```

mev_inspect/aave_liquidations.py
```diff
@@ -62,17 +64,21 @@ def get_aave_liquidations(
...
                     transaction_hash=trace.transaction_hash,
                     trace_address=trace.trace_address,
                     block_number=trace.block_number,
                 )
             )
+
     return liquidations
 
 
 def _get_liquidator_payback(
```
lukevs, line 77: the goal is always that someone can understand a function purely by its signature (name, inputs, and types)

It's not immediately clear what `Tuple[int, str]` is without reading the code

One option is to give it a name (ex: `_get_liquidator_payment_token_address_and_amount`)

Another option is to do what I think you said on the phone and actually return the full transfer (feels cleaner to me than this long name)

alembic/versions/205ce02374b3_add_received_collateral_address_to_liquidations.py
gheise: Gotcha

mev_inspect/aave_liquidations.py
gheise: returning the whole transfer will cause a variable return type. Rather have a better name and constant return type than variable return types

mev_inspect/aave_liquidations.py
lukevs: still here

mev_inspect/aave_liquidations.py
```diff
@@ -63,19 +71,29 @@ def get_aave_liquidations(
...
                     protocol=Protocol.aave,
                     received_amount=received_amount,
+                    received_token_address=received_token_address,
                     transaction_hash=trace.transaction_hash,
                     trace_address=trace.trace_address,
                     block_number=trace.block_number,
                 )
             )
+    print(liquidations)
     return liquidations
```
lukevs, line 81: the print is still here

mev_inspect/classifiers/specs/aave.py
```diff
@@ -1,10 +1,28 @@
...
+    TransferClassifier,
     LiquidationClassifier,
 )
+from mev_inspect.schemas.transfers import Transfer
+
+
+class AaveTransferClassifier(TransferClassifier):
+    @staticmethod
+    def get_transfer(trace: DecodedCallTrace) -> Transfer:
+        return Transfer(
```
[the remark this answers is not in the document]
gheise: too swag

lukevs (approved):

PR #90, opened 2021-10-06
Group classifying a trace as a `transfer` with the logic to decode a `Transfer` object

lukevs:
## How things are now

Let's say a new DEX comes out called LukeSwap

We'd like to add support for LukeSwap's swaps such that they write to the database (swaps table) and can be used in arbitrages etc

Currently this is a two step process:
1. Add a `ClassifierSpec` for LukeSwap and classify its `swap` function as a `Classification.swap`
2. Add support in the `get_swaps` function to check if the decoded abi / protocol was LukeSwap and if so decode it into a structured `Swap` object

Everyone does step 1, but it's easy to forget to do step 2 (I forgot for 0x!)

Additionally, as we add more and more protocols this `get_swaps` function will turn into a huge list of abi checks and function calls.

Ideally, the logic for classifying a trace at all (step 1) would live with and require instructions for decoding into a structured object (step 2)

## How things are after this PR

This PR changes `ClassifierSpec`s from specifying `Classification`s to specifying `Classifier`s

A `Classifier` currently is one of:
- `TransferClassifier`
- `SwapClassifier`
- `LiquidationClassifier`

All classifiers implement a `get_classification` to return the same enum we've been using

But now classifiers also implement functions to build their structured objects

For example `TransferClassifier`s must also implement
```
def get_transfer(trace: DecodedCallTrace) -> ERC20Transfer
```

See `ERC20TransferClassifier` and `WethTransferClassifier` for an example of implementing this

Implementing these structured object functions is sufficient for 1 and 2 above to work

And this means the logic for decoding lives alongside the other code associated with each specific protocol (rather than in a giant `get_transfers/swaps/etc` function)

Adding decode methods for `SwapClassifier` and `LiquidationClassifier` is left for future PRs

## Testing
- All existing tests pass
- Verified manually that existing blocks return the same transfer information

### DIFF
--- a/mev_inspect/schemas/classifiers.py
+++ b/mev_inspect/schemas/classifiers.py
@@ -11,7 +11,7 @@ class Classifier(ABC):
     @staticmethod
     @abstractmethod
     def get_classification() -> Classification:
-        pass
+        raise NotImplementedError()
 
 
 class TransferClassifier(Classifier):
@@ -22,7 +22,7 @@ class TransferClassifier(Classifier):
     @staticmethod
     @abstractmethod
     def get_transfer(trace: DecodedCallTrace) -> ERC20Transfer:
-        pass
+        raise NotImplementedError()
 
 
 class SwapClassifier(Classifier):


mev_inspect/transfers.py
```diff
@@ -18,9 +25,21 @@ def get_eth_transfers(traces: List[ClassifiedTrace]) -> List[EthTransfer]:
...
+    ] = {(spec.abi_name, spec.protocol): spec for spec in ALL_CLASSIFIER_SPECS}
+
     for trace in traces:
-        if trace.classification == Classification.transfer:
-            transfers.append(ERC20Transfer.from_trace(trace))
+        if not isinstance(trace, DecodedCallTrace):
+            continue
+
+        abi_name_and_protocol = (trace.abi_name, trace.protocol)
+        spec = specs_by_abi_name_and_protocol.get(abi_name_and_protocol)
```
lukevs, line 37: this logic is a little gross and will be moved to a shared function as swap and liquidate are added

mev_inspect/schemas/classifiers.py
```diff
@@ -0,0 +1,44 @@
...
+    @staticmethod
+    def get_classification() -> Classification:
+        return Classification.liquidate
+
+
+class ClassifierSpec(BaseModel):
+    abi_name: str
+    protocol: Optional[Protocol] = None
+    valid_contract_addresses: Optional[List[str]] = None
+    classifiers: Dict[str, Type[Classifier]] = {}
```
lukevs, line 44: `Type[Classifier]` means we accept the class itself (ex: `WethTransferClassifier`) rather than an instance of the class (ex: `WethTransferClassifier()`) to further enforce that these implement only static methods

bertmiller:
This all follows to me, we just need to get it on the docs for other contributors to be able to understand

What do you need to get this PR over the line?

PR #91, opened 2021-10-07
Classifiers: add 1inch Limit Order Protocol

come-maiz:


elopio:
Thanks a lot @taarushv! <3

I've been debugging this, and it is blocked because there is a bug when we get the selector of a function that has a tuple. I'm having a lot of fun digging into this :)

mev_inspect/classifiers/specs/one_inch.py
```diff
@@ -0,0 +1,37 @@
...
+
+    @staticmethod
+    def get_swap_recipient(trace: DecodedCallTrace) -> str:
+        return trace.from_address
+
+
+FILL_ORDER_SIGNATURE = (
+    "fillOrder((uint256,address,address,bytes,bytes,bytes,bytes,bytes,bytes,bytes),"
+    "bytes,uint256,uint256,uint256)"
+)
```
lukevs, line 21: wow 1inch - beefy signatures

mev_inspect/classifiers/specs/one_inch.py
```diff
@@ -0,0 +1,37 @@
...
+
+ONE_INCH_LIMIT_ORDER_PROTOCOL_SPEC = ClassifierSpec(
+    abi_name="1inchLimitOrderProtocol",
+    protocol=Protocol.one_inch,
+    classifications={
+        FILL_ORDER_SIGNATURE: OneInchFillOrderProtocolSwapClassifier,
+    },
+)
+
+ONE_INCH_CLASSIFIER_SPECS = [ONE_INCH_LIMIT_ORDER_PROTOCOL_SPEC]
```
lukevs, line 37: one last thing for this to work is adding it to `mev_inspect/classifiers/__init__.py` `ALL_CLASSIFIER_SPECS`

elopio: done

lukevs (approved):

elopio:
@lukevs there was one merge problem, but now it's all good. I'm back again at my pool question. This transaction doesn't have a pool. Does it mean that it's not a swap?

lukevs:
@elopio this is something we ran into w/ 0x as well. thinking we might want to rename that field to just contract_address

elopio:
pool and contract address, both sound good to me. But in this case there is no transfer to the contract, just a direct exchange. As far as I understand this function.

PR #94, opened 2021-10-11
decode: collapse tuples

### LINKED ISSUE #93: Fail to classify an ABI function with a tuple argument
When the ABI is decoded, it just collects all the types of the arguments:

https://github.com/flashbots/mev-inspect-py/blob/b6fc27b3f670e9933c9bdf9e2274d34b845c89e5/mev_inspect/decode.py#L29

This results on the wrong selector being calculated, and then a failure to classify any function that has a tuple as argument.

come-maiz:
Closes #93.

### DIFF
--- a/mev_inspect/decode.py
+++ b/mev_inspect/decode.py
@@ -1,5 +1,6 @@
-import eth_utils.abi
 from typing import Dict, Optional
+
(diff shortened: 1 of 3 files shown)


elopio:
done!

PR #96, opened 2021-10-11
Remove ETH / ERC20 transfer distinction

lukevs:
Currently we only support ERC20s when storing transfers

For some new protocols we want to support (ex: curve) swaps occur from ETH to ERC20 tokens directly.

This PR adds support for ETH transfers by:
- Removing Eth/ERC20Transfer and instead sharing a single Transfer object
- Representing EthTransfers using the `0xeeeeeeeee...` token address (a pattern set by Bancor, Kyber, etc)

### DIFF
--- a/mev_inspect/transfers.py
+++ b/mev_inspect/transfers.py
@@ -33,15 +33,15 @@ def get_eth_transfers(traces: List[ClassifiedTrace]) -> List[Transfer]:
 
 def get_transfer(trace: ClassifiedTrace) -> Optional[Transfer]:
     if trace.value is not None and trace.value > 0:
-        return _build_eth_transfer(trace)
+        return _get_eth_transfer(trace)
 
     if isinstance(trace, DecodedCallTrace):
-        return _build_erc20_transfer(trace)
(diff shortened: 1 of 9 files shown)


mev_inspect/transfers.py
```diff
@@ -3,40 +3,57 @@
...
+        transfer
+        for transfer in transfers
+        if transfer.token_address == ETH_TOKEN_ADDRESS
+    ]
 
-    return transfers
 
+def get_transfer(trace: ClassifiedTrace) -> Optional[Transfer]:
+    if trace.value is not None and trace.value > 0:
+        return _get_eth_transfer(trace)
```
lukevs, line 36: Note to self - this needs to change

This will represent any function call that's payable as a transfer. Should only be true transfers?

bertmiller: I'm not sure you can distinguish between the two. e.g. an ETH -> stETH swap on Curve is a payable function.

bertmiller (approved):
LGTM

PR #104, opened 2021-10-14
Support for Cream markets + tests

taarushv:
* Added ABI again (despite sharing with compound/being a fork) because trace classifier looks up ABI based on protocol tag (which I needed to be distinct, for the inspector). 
* Added a `has_liquidations` trace check to avoid network calls unless there is a liquidation (Cream has 90+ markets, looking up underlying for each at every block would've slowed things down)
* Made sure the inspector shared the same code to make it easy to add other forks
* Test case for a cream liquidation
* Replaced WETH for 0xeee... as a default stand in for ETH


mev_inspect/compound_liquidations.py
```diff
@@ -54,7 +80,18 @@ def get_compound_liquidations(
...
+                underlying_markets = collateral_by_c_token_address
+            elif trace.protocol == Protocol.cream:
+                underlying_markets = collateral_by_cr_token_address
+            else:
+                underlying_markets = {}
+
+            if (
+                seize_trace is not None
+                and seize_trace.inputs is not None
+                and underlying_markets is not {}
```
lukevs, line 93: `is` compares object identity and should only be used for `None` checks
```
>>> {} is {}
False
```

should be `len(underlying_markets) == 0`

mev_inspect/compound_liquidations.py
```diff
@@ -35,8 +37,29 @@ def fetch_all_comp_markets(w3: Web3) -> Dict[str, str]:
...
+    cr_token_mapping = {}
+    cream_comptroller_abi = get_raw_abi("Comptroller", Protocol.compound_v2)
+    comptroller_instance = w3.eth.contract(
+        address=CREAM_COMPTROLLER_ADDRESS, abi=cream_comptroller_abi
+    )
+    markets = comptroller_instance.functions.getAllMarkets().call()
+    cream_crtoken_abi = get_raw_abi("CToken", Protocol.compound_v2)
+    for cr_token in markets:
+        # make an exception for cETH (as it has no .underlying())
+        if cr_token != CREAM_CR_ETHER:
```
lukevs, line 50: How does this affect things downstream? Will we get a `KeyError` for `CREAM_CR_ETHER` when trying to lookup the collateral address?

mev_inspect/liquidations.py
```diff
@@ -4,16 +4,33 @@
...
-    aave_liquidations = get_aave_liquidations(classified_traces)
-    comp_markets = fetch_all_comp_markets(w3)
-    compound_liquidations = get_compound_liquidations(classified_traces, comp_markets)
-    return aave_liquidations + compound_liquidations
+    # to avoid contract calls to fetch comp/cream markets
+    # unless there is a liquidation
+    has_liquidations = False
+    for classified_trace in classified_traces:
+        if classified_trace.classification == Classification.liquidate:
+            has_liquidations = True
```
lukevs, line 25: this bit would be good as a helper function
```
if has_liquidations(classified_traces)
```

PR #112, opened 2021-10-27
Geth support for inspect

supragya:
Added support for geth nodes by introducing a translation layer at RPC call sites. Translates geth traces and geth receipts to parity like traces and parity like receipts. Works for analyzing arbitrages on polygon chain.

### DIFF
--- a/mev_inspect/block.py
+++ b/mev_inspect/block.py
@@ -17,6 +17,7 @@ _calltype_mapping = {
     "CALL": "call",
     "DELEGATECALL": "delegateCall",
     "CREATE": "create",
+    "CREATE2": "create2",
     "SUICIDE": "suicide",
     "REWARD": "reward",
 }
--- a/README.md
+++ b/README.md
@@ -66,19 +66,19 @@ On first startup, you'll need to apply database migrations with:
 ### Inspect a single block
 
 Inspecting block [12914944](https://twitter.com/mevalphaleak/status/1420416437575901185):
-**Note**: Add `--geth` at the end if RPC_URL points to a geth / geth like node.
+**Note**: Add `geth` at the end instead of `parity` if RPC_URL points to a geth / geth like node.
 
 ```
-./mev inspect 12914944
+./mev inspect 12914944 parity
 ```
 
 ### Inspect many blocks
 
 Inspecting blocks 12914944 to 12914954:
-**Note**: Add `--geth` at the end if RPC_URL points to a geth / geth like node.
+**Note**: Add `geth` at the end instead of `parity` if RPC_URL points to a geth / geth like node.
 
 ```
-./mev inspect-many 12914944 12914954
+./mev inspect-many 12914944 12914954 parity
 ```
 
 ### Inspect all incoming blocks
--- a/mev
+++ b/mev
@@ -37,15 +37,17 @@ case "$1" in
 	;;
   inspect)
         block_number=$2
+        rpc_type=$3
         echo "Inspecting block $block_number"
-        kubectl exec -ti deploy/mev-inspect -- poetry run inspect-block $block_number
+        kubectl exec -ti deploy/mev-inspect -- poetry run inspect-block $block_number --type $rpc_type
 	;;
   inspect-many)
         start_block_number=$2
         end_block_number=$3
+        rpc_type=$4
         echo "Inspecting from block $start_block_number to $end_block_number"
         kubectl exec -ti deploy/mev-inspect -- \
-            poetry run inspect-many-blocks $start_block_number $end_block_number
+            poetry run inspect-many-blocks $start_block_number $end_block_number --type $rpc_type
 	;;
   test)
         echo "Running tests"


cli.py
```diff
@@ -26,12 +27,15 @@ def cli():
 @click.argument("block_number", type=int)
 @click.option("--rpc", default=lambda: os.environ.get(RPC_URL_ENV, ""))
 @click.option("--cache/--no-cache", default=True)
-def inspect_block_command(block_number: int, rpc: str, cache: bool):
+@click.option("--geth/--no-geth", default=False)
```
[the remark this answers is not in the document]
shekhirin: yeah, it seems to be possible to check whether the first RPC query failed or not and decide which method to use based on that. Then, if RPC endpoint client changes and request fails again, we could do the same thing just checking the response code every time.

OpenEthereum also supports block tracing btw: https://openethereum.github.io/JSONRPC-trace-module#trace_block

mev_inspect/block.py
```diff
@@ -36,43 +39,64 @@ async def create_from_block_number(
...
-            Receipt(**receipt) for receipt in receipts_json["result"]
-        ]
-        traces = [Trace(**trace_json) for trace_json in traces_json["result"]]
-    except KeyError as e:
-        logger.warning(
-            f"Failed to create objects from block: {block_number}: {e}, retrying: {retries + 1} / 3"
+async def _fetch_block(
+    w3, base_provider, geth: bool, block_number: int, retries: int = 0
+) -> Block:
+    if not geth:
```
lukevs, line 50: mind breaking into two functions? `_fetch_block_parity` / `_fetch_block_geth` and have the if switch between

mev_inspect/block.py
```diff
@@ -36,43 +39,64 @@ async def create_from_block_number(
...
+            if retries < 3:
+                await asyncio.sleep(5)
+                return await _fetch_block(
+                    w3, base_provider, geth, block_number, retries
+                )
+            else:
+                raise
+    else:
+        # print(block_number)
+        block_json = await asyncio.gather(w3.eth.get_block(block_number))
```
lukevs, line 84: can remove gather here since it's just one entry and await instead

mev_inspect/block.py
```diff
@@ -191,3 +215,140 @@ def get_transaction_hashes(calls: List[Trace]) -> List[str]:
...
+    for idx, tx_receipt in enumerate(json_decoded_receipts):
+        if tx_receipt != None:
+            results.append(unwrap_tx_receipt_for_parity(block_json, idx, tx_receipt))
+    return results
+
+
+def unwrap_tx_receipt_for_parity(block_json, tx_pos_in_block, tx_receipt) -> Receipt:
+    try:
+        if tx_pos_in_block != int(tx_receipt["transactionIndex"], 16):
+            print(
```
lukevs, line 336: logger.info instead of print so we can manager where the output goes

mev_inspect/block.py
```diff
@@ -191,3 +215,140 @@ def get_transaction_hashes(calls: List[Trace]) -> List[str]:
...
+            transaction_hash=tx_receipt["transactionHash"],
+            transaction_index=tx_pos_in_block,
+            gas_used=tx_receipt["gasUsed"],
+            effective_gas_price=tx_receipt["effectiveGasPrice"],
+            cumulative_gas_used=tx_receipt["cumulativeGasUsed"],
+            to=tx_receipt["to"],
+        )
+
+    except Exception as e:
+        print("error while decoding receipt", tx_receipt, e)
```
lukevs, line 352: What causes this to get hit?

mev_inspect/block.py
```diff
@@ -191,3 +215,140 @@ def get_transaction_hashes(calls: List[Trace]) -> List[str]:
...
+                result=result_dict,
+                subtraces=len(tx_trace["calls"]) if "calls" in tx_trace.keys() else 0,
+                trace_address=position,
+                transaction_hash=block_json["transactions"][tx_pos_in_block].hex(),
+                transaction_position=tx_pos_in_block,
+                type=TraceType(_calltype_mapping[tx_trace["type"]]),
+            )
+        )
+    except Exception:
+        return []
```
lukevs, line 283: what causes this to get hit

mev_inspect/block.py
```diff
@@ -191,3 +215,140 @@ def get_transaction_hashes(calls: List[Trace]) -> List[str]:
...
+    return response_list
+
+
+async def geth_get_tx_receipts_task(session, endpoint_uri, tx):
+    data = {
+        "jsonrpc": "2.0",
+        "id": "0",
+        "method": "eth_getTransactionReceipt",
+        "params": [tx.hex()],
+    }
```
lukevs, line 301: Instead of making a custom session, does `base_provider.make_request("eth_getTransactionReceipt", [tx.hex()]),` work?

That would allow us to keep middleware like our retry logic

mev_inspect/block.py
```diff
@@ -191,3 +215,140 @@ def get_transaction_hashes(calls: List[Trace]) -> List[str]:
...
+
+
+async def geth_get_tx_receipts_async(endpoint_uri, transactions):
+    geth_tx_receipts = []
+    async with aiohttp.ClientSession() as session:
+        tasks = [
+            asyncio.create_task(geth_get_tx_receipts_task(session, endpoint_uri, tx))
+            for tx in transactions
+        ]
+        geth_tx_receipts = await asyncio.gather(*tasks)
```
lukevs, line 315:
```
tasks = [single_task()]
result = await asyncio.gather(*tasks)
```

I believe is equivalent to just
```
result = await single_task()
```

mev_inspect/block.py
```diff
@@ -36,43 +39,64 @@ async def create_from_block_number(
...
+                )
+            else:
+                raise
+    else:
+        # print(block_number)
+        block_json = await asyncio.gather(w3.eth.get_block(block_number))
+        traces = await geth_get_tx_traces_parity_format(base_provider, block_json[0])
+        geth_tx_receipts = await geth_get_tx_receipts_async(
+            base_provider.endpoint_uri, block_json[0]["transactions"]
+        )
```
lukevs, line 88: these can be grouped with a gather to make in parallel

mev_inspect/block.py
```diff
@@ -191,3 +215,140 @@ def get_transaction_hashes(calls: List[Trace]) -> List[str]:
...
+        "debug_traceBlockByHash", [block_hash.hex(), {"tracer": "callTracer"}]
+    )
+    return block_trace
+
+
+def unwrap_tx_trace_for_parity(
+    block_json, tx_pos_in_block, tx_trace, position=[]
+) -> List[Trace]:
+    response_list = []
+    _calltype_mapping = {
```
lukevs, line 248: Move to a constant up top

mev_inspect/geth_poa_middleware.py
```diff
@@ -0,0 +1,97 @@
+from typing import (
```
lukevs, line 1: Want to call out in a docstring or comment up top that this is from web3.py until they add support?

Ideally they support this at some point and we move to that

cli.py
```diff
@@ -38,7 +40,7 @@ async def fetch_block_command(block_number: int, rpc: str):
     inspect_db_session = get_inspect_session()
     trace_db_session = get_trace_session()
 
-    inspector = MEVInspector(rpc, inspect_db_session, trace_db_session)
+    inspector = MEVInspector(rpc, inspect_db_session, trace_db_session, False)
```
lukevs, line 43: Should support here too. Should be able to use the functions already needed for the other stuff

cli.py
```diff
@@ -21,12 +21,14 @@ def cli():
 @cli.command()
 @click.argument("block_number", type=int)
 @click.option("--rpc", default=lambda: os.environ.get(RPC_URL_ENV, ""))
+@click.option("--geth/--no-geth", default=False)
 @coro
-async def inspect_block_command(block_number: int, rpc: str):
+async def inspect_block_command(block_number: int, rpc: str, geth: bool):
```
lukevs, line 26: Mentioned this in our channel a bit ago - I think it would be better to pass around `rpc_type` where it's an enum of `geth` or `parity` to make geth less "exceptional" and just make it one of our supported RPC types.

Then the code itself would read nicer as well
```
if rpc_type == RPCType.geth:
    return _fetch_block_geth(...)
elif rpc_type == RPCType.parity:
    return _fetch_block_parity(...)
...
```

mev_inspect/block.py
supragya: We haven't ever seen it get hit. I'll leave it to the caller to do try..catch for now

mev_inspect/block.py
supragya: from my experience, `CREATE2` and `SUICIDE`. Both not used in analysis

mev_inspect/block.py
supragya: that's list of tasks right? `for tx in transactions`

mev_inspect/block.py
supragya: Have seen extra loads while tracing is being done hamper tracer in polygon. Sometime the tracer runs out of time (internal timeout) and errors. Better to keep separate.

mev_inspect/block.py
```diff
@@ -36,43 +39,64 @@ async def create_from_block_number(
...
+                raise
+    else:
+        # print(block_number)
+        block_json = await asyncio.gather(w3.eth.get_block(block_number))
+        traces = await geth_get_tx_traces_parity_format(base_provider, block_json[0])
+        geth_tx_receipts = await geth_get_tx_receipts_async(
+            base_provider.endpoint_uri, block_json[0]["transactions"]
+        )
+        receipts = geth_receipts_translator(block_json[0], geth_tx_receipts)
+        base_fee_per_gas = 0
```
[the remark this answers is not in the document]
supragya: Should be, however, we don't have such in polygon. Could be taken up in a separate PR/GH issue.

mev_inspect/block.py
lukevs: If that's the case, want to move this to the same as you have for STATICCALL where it checks for it explicitly?

It's better on our side to have it just blow up with an exception for now. We'd rather know if we're missing cases.

mev_inspect/block.py
lukevs: Whoops! yes - my bad. Missed `for tx in transactions`

mev_inspect/block.py
supragya: Doesn't seem to work without asyncio.gather

mev_inspect/block.py
supragya: 1. Added `create2`
2. Added `logger` warning.
Haven't yet made it to throw exception when stuff happens since it helps us back-fill in "best effort" form. Yet, person who wants to work on this further for translation to be "exact" can remove this. Can add a comment around here though. What do you think? @lukevs

mev_inspect/utils.py
```diff
@@ -1,5 +1,11 @@
+from enum import Enum
 from hexbytes._utils import hexstr_to_bytes
 
 
+class RPCType(Enum):
+    parity = 0
+    geth = 1
```
lukevs, line 7: Elsewhere we do like:
```
class RPCType(Enum):
    parity = "parity"
    geth = "geth"
```

which then allows you to in click etc do
```
str_rpc_type = "parity"
rpc_type = RPCType(str_rpc_type)
```
instead of custom conversion logic

CodeForcer:
Is there any appetite for getting this merged into main? I think it's a really crucial addition in order to make inspect multichain

PR #115, opened 2021-10-28
Add 0x classifiers

gheise:
## Add coverage for 0x

Currently, many 0x smart contract ABI's are implemented without classification. The goal here is to implement classifications for as many function as possible.


mev_inspect/classifiers/specs/zero_ex.py
```diff
@@ -3,9 +3,17 @@
...
     ClassifierSpec,
+    DecodedCallTrace,
+    SwapClassifier,
 )
 
 
+class ZeroXSwapClassifier(SwapClassifier):
+    @staticmethod
+    def get_swap_recipient(trace: DecodedCallTrace) -> str:
+        return trace.inputs["recipient"]
```
lukevs, line 14: there's aways a recipient specified or does it ever default to from address?

gheise: this was actually incorrect, but i've corrected it

PR #117, opened 2021-11-02
Reindex swaps by block number

lukevs:
On inspection, we delete all entries by block number for each table before recreating

Swaps are the only table without an index on block number, which makes their deletes slow as the table gets bigger

This PR reindexes the primary key for swaps by block number to make their deletion fast as well


bertmiller (approved):
LGTM

PR #125, opened 2021-11-10
Compound backfilling/removed network calls

taarushv:
This PR:
* Creates a table `ctoken_underlying` to store cTokens/Underlying collateral addresses/symbol (to avoid network calls during classified traces inspection, queries DB instead)
* Adds a command `./mev backfill-lending-markets` to trigger backfill
* Adds a `check_for_updates` helper and a `new-market` classifier spec to identify/store new markets at the beginning of every block
* Test case to make sure we identify new markets (by using the block in which MKR was added to compound)

After `./mev exec alembic upgrade head` and `./mev backfill-lending-markets` tested with the following to make sure it picks up compound/cream liquidations and adds to db: 
* `./mev inspect 13234998`
* ` ./mev inspect 12674514`

Edit: PR has since pivoted to removing a column and addressing lookups via join

### DIFF
--- a/mev_inspect/compound_liquidations.py
+++ b/mev_inspect/compound_liquidations.py
@@ -1,5 +1,4 @@
-from typing import Dict, List, Optional
-from web3 import Web3
+from typing import List, Optional
 
 from mev_inspect.traces import get_child_traces
 from mev_inspect.schemas.traces import (
@@ -9,44 +8,15 @@ from mev_inspect.schemas.traces import (
 )
 
 from mev_inspect.schemas.liquidations import Liquidation
-from mev_inspect.abi import get_raw_abi
-from mev_inspect.transfers import ETH_TOKEN_ADDRESS
 
 V2_COMPTROLLER_ADDRESS = "0x3d9819210A31b4961b30EF54bE2aeD79B9c9Cd3B"
 V2_C_ETHER = "0x4Ddc2D193948926D02f9B1fE9e1daa0718270ED5"
 CREAM_COMPTROLLER_ADDRESS = "0x3d5BC3c8d13dcB8bF317092d84783c2697AE9258"
 CREAM_CR_ETHER = "0xD06527D5e56A3495252A528C4987003b712860eE"
 
-# helper, only queried once in the beginning (inspect_block)
-def fetch_all_underlying_markets(w3: Web3, protocol: Protocol) -> Dict[str, str]:
-    if protocol == Protocol.compound_v2:
-        c_ether = V2_C_ETHER
-        address = V2_COMPTROLLER_ADDRESS
-    elif protocol == Protocol.cream:
-        c_ether = CREAM_CR_ETHER
-        address = CREAM_COMPTROLLER_ADDRESS
-    else:
-        raise ValueError(f"No Comptroller found for {protocol}")
-    token_mapping = {}
-    comptroller_abi = get_raw_abi("Comptroller", Protocol.compound_v2)
-    comptroller_instance = w3.eth.contract(address=address, abi=comptroller_abi)
-    markets = comptroller_instance.functions.getAllMarkets().call()
-    token_abi = get_raw_abi("CToken", Protocol.compound_v2)
-    for token in markets:
-        # make an exception for cETH (as it has no .underlying())
-        if token != c_ether:
-            token_instance = w3.eth.contract(address=token, abi=token_abi)
-            underlying_token = token_instance.functions.underlying().call()
-            token_mapping[
-                token.lower()
-            ] = underlying_token.lower()  # make k:v lowercase for consistancy
-    return token_mapping
-
 
 def get_compound_liquidations(
     traces: List[ClassifiedTrace],
-    collateral_by_c_token_address: Dict[str, str],
-    collateral_by_cr_token_address: Dict[str, str],
 ) -> List[Liquidation]:
 
     """Inspect list of classified traces and identify liquidation"""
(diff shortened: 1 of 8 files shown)


alembic/versions/7a34fb3655a2_add_ctoken_table.py
```diff
@@ -0,0 +1,28 @@
...
+down_revision = '205ce02374b3'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        "ctoken_underlying",
+        sa.Column("ctoken_address", sa.String(256), nullable=False, primary_key=True),
+        sa.Column("ctoken_symbol", sa.String(256), nullable=False),
```
lukevs, line 23: it looks like this is only written to not read

I think better to leave out in that case, since it requires an additional fetch

backfill-lending-markets.py
```diff
@@ -0,0 +1,49 @@
+import os
+from web3 import Web3
+from mev_inspect.compound_liquidations import fetch_all_underlying_markets, fetch_token_symbol
+from mev_inspect.crud.liquidations import write_ctoken_markets
+from mev_inspect.schemas.liquidations import CTokenMarket
+from mev_inspect.schemas.traces import Protocol
+from mev_inspect.db import get_inspect_session
+
+def store_markets():
```
lukevs, line 9: I think better for this to live in `cli.py` instead of its own script

`backfill.py` is really a bash script, but I did it in python because i'm bad at bash scripting lol

backfill-lending-markets.py
```diff
@@ -0,0 +1,49 @@
...
+    rpc = os.getenv("RPC_URL")
+    if rpc is None:
+        raise RuntimeError("Missing env variable RPC_URL, required for lending markets backfilling")
+    
+    w3 = Web3(Web3.HTTPProvider(rpc))
+    print("Fetching COMP markets...")
```
lukevs, line 15: in cli, better to use the logger here instead of prints

mev_inspect/utils.py
```diff
@@ -1,5 +1,42 @@
...
+from mev_inspect.schemas.traces import Classification, ClassifiedTrace, Protocol
+
+COMPOUND_COMPTROLLER_ADDRESS = "0x95b4ef2869ebd94beb4eee400a99824bf5dc325b"
 
 
 def hex_to_int(value: str) -> int:
     return int.from_bytes(hexstr_to_bytes(value), byteorder="big")
+
+
+def check_for_updates(classified_traces: List[ClassifiedTrace], w3: Web3, db_session):
```
lukevs, line 18: `utils.py` should only be for simple shared utils - would be better to have this as in `mev_inspect/lending.py`

lukevs: This should have a more specific function name as well

mev_inspect/crud/liquidations.py
```diff
@@ -29,3 +29,24 @@ def write_liquidations(
 
     db_session.bulk_save_objects(models)
     db_session.commit()
+
+
+def fetch_all_ctoken_markets(db_session) -> Dict[str, str]:
```
lukevs, line 34: `find_` for DB queries

mev_inspect/crud/liquidations.py
```diff
@@ -29,3 +29,24 @@ def write_liquidations(
...
+    result = db_session.query(CTokenModel).all()
+    markets: Dict[str, str] = {}
+    if result == []:
+        return markets
+    else:
+        for market in result:
+            markets[market.ctoken_address] = market.underlying_token_address
+        return markets
```
lukevs, line 42: This is the same as not having the if which I think is clearer
```
markets: Dict[str, str] = {}
for market in result:
    markets[market.ctoken_address] = market.underlying_token_address
return markets
```

mev_inspect/liquidations.py
```diff
@@ -17,7 +19,17 @@ def has_liquidations(classified_traces: List[ClassifiedTrace]) -> bool:
 
 
 def get_liquidations(
+    db_session,
     classified_traces: List[ClassifiedTrace],
 ) -> List[Liquidation]:
     aave_liquidations = get_aave_liquidations(classified_traces)
+    c_token_underlying_markets = fetch_all_ctoken_markets(db_session)
```
lukevs, line 26: can this be moved outside so we don't redo the DB find on each iteration?

tests/test_ctoken_backfill.py
```diff
@@ -0,0 +1,16 @@
...
+    # MKR was added to COMP in this block
+    block_number = 12949662
+    block = load_test_block(block_number)
+    trace_classifier = TraceClassifier()
+    classified_traces = trace_classifier.classify(block.traces)
+    has_new_market = False
+    for trace in classified_traces:
+        if(trace.classification == Classification.new_market and trace.inputs['cToken'] == '0x95b4ef2869ebd94beb4eee400a99824bf5dc325b'):
+            has_new_market = True
+    assert has_new_market == True
```
lukevs, line 15: can be just
```
assert has_new_market
```

mev_inspect/utils.py
lukevs: removing the write is also good because it means you can write a static test for this

mev_inspect/compound_liquidations.py
```diff
@@ -101,7 +103,7 @@ def get_compound_liquidations(
                     liquidations.append(
                         Liquidation(
                             liquidated_user=trace.inputs["borrower"],
-                            collateral_token_address=underlying_markets[
+                            collateral_token_address=c_token_underlying_markets[
```
lukevs, line 106: There's a state here where someone:
- Doesn't run `./mev backfill-lending-markets`
- Encounters a compound liquidation

Which will cause a `KeyError`

mev_inspect/compound_liquidations.py
```diff
@@ -43,10 +43,17 @@ def fetch_all_underlying_markets(w3: Web3, protocol: Protocol) -> Dict[str, str]
     return token_mapping
 
 
+def fetch_token_symbol(w3: Web3, token_address: str) -> str:
```
lukevs, line 46: This isn't used anymore

cli.py
```diff
@@ -94,6 +99,43 @@ async def fetch_all_prices():
     write_prices(inspect_db_session, prices)
 
 
+@cli.command()
+@coro
+async def store_markets():
+    rpc = os.getenv("RPC_URL")
```
lukevs, line 105: Could you use the same pattern as the other commands - ex:
```
@click.option("--rpc", default=lambda: os.environ.get(RPC_URL_ENV, ""))
```

lukevs (approved):

PR #130, opened 2021-11-17
Implement swap classifiers

gheise:
## Extend the SwapClassifier
The goal here is to extend the SwapClassifier. Rather than simply fetching the recipient address, the SwapClassifier should actually create the swap. This makes the swap logic protocol specific, which differentiates a swap between exchanges and higher order protocols, and will facilitate the implementation of future coverage.

Note: Included is a necessary change to ./mev file to run commands on some systems. This should not cause any problems with the current support, but is still up for review.


mev
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
```
elopio, line 1: I suggest to make this change in a separate branch.

mev_inspect/classifiers/utils.py
```diff
@@ -0,0 +1,89 @@
...
+def create_swap_from_transfers(
+    trace: DecodedCallTrace,
+    recipient_address: str,
+    prior_transfers: List[Transfer],
+    child_transfers: List[Transfer],
+) -> Optional[Swap]:
+    pool_address = trace.to_address
+
+    if recipient_address is None:
+        return None
```
elopio, line 18: why not raise an exception? This seems like a silent failure.

mev_inspect/classifiers/utils.py
```diff
@@ -0,0 +1,89 @@
...
+        transfers_to_pool = [_build_eth_transfer(trace)]
+
+    if len(transfers_to_pool) == 0:
+        transfers_to_pool = _filter_transfers(prior_transfers, to_address=pool_address)
+
+    if len(transfers_to_pool) == 0:
+        transfers_to_pool = _filter_transfers(child_transfers, to_address=pool_address)
+
+    if len(transfers_to_pool) == 0:
+        return None
```
elopio, line 32: These 3 conditionals I think that deserve a comment. It's not clear what's happenineth_sendPrivateTransactiong. First we try to find prior transfers, if it fails, we try to find child transfers, is that right?

This prior transfers thing is new, right? I don't understand how to find them. It would be nice to have a documented guide of how to make a new swap classifier.

And again, I think it would be clearer to raise an exception instead of returning None. Like exception! the trace is not a swap because it doesn't have transfers to the pool.

mev_inspect/classifiers/specs/balancer.py
```diff
@@ -6,15 +9,25 @@
     ClassifierSpec,
     SwapClassifier,
 )
-
+from mev_inspect.classifiers.utils import create_swap_from_transfers
```
lukevs, line 12: "utils" is a placeholder word

how about `mev_inspect.classifiers.swaps`?

mev_inspect/classifiers/specs/uniswap.py
```diff
@@ -14,20 +18,40 @@
...
 class UniswapV3SwapClassifier(SwapClassifier):
     @staticmethod
-    def get_swap_recipient(trace: DecodedCallTrace) -> str:
+    def parse_swap(
+        trace: DecodedCallTrace,
+        prior_transfers: List[Transfer],
+        child_transfers: List[Transfer],
+    ) -> Optional[Swap]:
+
         if trace.inputs is not None and "recipient" in trace.inputs:
```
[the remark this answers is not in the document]
lukevs: In which case this can all turn into
```
recipient_address = trace.inputs.get("recipient", trace.from_address)
```

mev_inspect/classifiers/specs/uniswap.py
```diff
@@ -14,20 +18,40 @@
...
-    def get_swap_recipient(trace: DecodedCallTrace) -> str:
+    def parse_swap(
+        trace: DecodedCallTrace,
+        prior_transfers: List[Transfer],
+        child_transfers: List[Transfer],
+    ) -> Optional[Swap]:
+
         if trace.inputs is not None and "to" in trace.inputs:
-            return trace.inputs["to"]
+            recipient_address = trace.inputs["to"]
```
lukevs, line 47: same comment as in V3

mev_inspect/classifiers/utils.py
lukevs: can recipient address be None? you have it type annotated as str not Optional[str]

gheise: the recipient address can't be none as it is passed in parse_swap as an attribute of a DecodedCallTrace. This logic was already here, which i did not intend to alter and move in the same PR. This only moves the existing swap logic into the swap classifiers

mev_inspect/classifiers/specs/balancer.py
gheise: Thanks, great feedback

mev_inspect/classifiers/specs/uniswap.py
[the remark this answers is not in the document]
gheise: inputs can't be none, agreed, but should still check for the presence of the recipient, otherwise it'll throw KeyError

mev_inspect/classifiers/utils.py
gheise: I agree this would make things much clearer. The goal here was to port the existing logic from swaps into the classifiers, no logic changes. I think this would be better addressed in its own PR, then we can throw for swaps, transfers, liquidations, etc, and address this properly as well as maintain this concise. Thoughts?

mev_inspect/classifiers/specs/uniswap.py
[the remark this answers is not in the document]
lukevs: `get` does that

it says "if recipient is present, get it's value - otherwise use this default"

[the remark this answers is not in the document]
gheise: Cool - will add this!

mev_inspect/classifiers/utils.py
elopio: I agree.

lukevs (approved):

PR #134, opened 2021-11-18
Add support for fetching prices from coinbase and storing

lukevs:
Run with
```
./mev prices fetch-all
```

Which pulls down all prices for all supported assets (just WETH / ETH for now) and writes them to the database

Duplicates are ignored on re-insert so we can re-run this multiple times and only insert new


bertmiller (approved):
Same feedback as #142

PR #141, opened 2021-11-22
Add classifier helpers

gheise:
## Add a helper file with functions used in swap classifiers

-  Rename existing swaps file to helpers, better naming

### DIFF
--- a/mev_inspect/classifiers/helpers.py
+++ b/mev_inspect/classifiers/helpers.py
@@ -84,14 +84,3 @@ def _filter_transfers(
         filtered_transfers.append(transfer)
 
     return filtered_transfers
-
-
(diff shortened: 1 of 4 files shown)


mev_inspect/classifiers/helpers.py
```diff
@@ -84,3 +84,14 @@ def _filter_transfers(
         filtered_transfers.append(transfer)
 
     return filtered_transfers
+
+
+def get_amount_transferred_to_address(
+    address: str, transfers: List[Transfer]
+) -> Optional[int]:
+
+    for transfer in transfers:
```
lukevs, line 93: what if there are multiple transfers either of the same or different tokens

lukevs (approved):

PR #142, opened 2021-11-22
Add cron job to fetch prices

lukevs:
Adds a Kubernetes `CronJob` to run fetch prices every hour at minute 0

We can always update in the future to run more frequently but this is sufficient for now


Tiltfile
```diff
@@ -30,8 +29,7 @@ k8s_yaml(secret_from_dict("mev-inspect-db-credentials", inputs = {
 #     "host": "trace-db-postgresql",
 # }))
 
-docker_build_with_restart("mev-inspect-py", ".",
-    entrypoint="/app/entrypoint.sh",
+docker_build("mev-inspect-py", ".",
```
lukevs, line 32: entrypoint wasn't working with args passed

this just means we need a full restart on loop updates

mev_inspect/coinbase.py
```diff
@@ -0,0 +1,25 @@
+import aiohttp
+
+from mev_inspect.classifiers.specs.weth import WETH_ADDRESS
+from mev_inspect.schemas.transfers import ETH_TOKEN_ADDRESS
+from mev_inspect.schemas.coinbase import CoinbasePrices, CoinbasePricesResponse
+
+
+COINBASE_API_BASE = "https://www.coinbase.com/api/v2"
+COINBASE_TOKEN_NAMES = {
+    WETH_ADDRESS: "weth",
```
bertmiller, line 10: Why are these called addresses? A bit confusing because they aren't eth addresses

bertmiller (approved):
lgtm although you know cron/kube/etc much better than I do

mev_inspect/coinbase.py
lukevs: The dict itself is address => "coinbase name". The `WETH_ADDRESS` is an address

I should probably rename to `COINBASE_TOKEN_NAME_BY_ADDRESS`

Thoughts?

bertmiller: That makes sense to me

lukevs: Updated - marking resolves

PR #143, opened 2021-11-23
Add support for 0x orderbook

gheise:
## Support for 0x
- The goal here is to implement support for 0x's orderbook. 
- Since 0x's Uniswap feature is already supported through the Uniswap classifiers, the orderbook is the next big MEV-related feature of interest. 
- Included is the 0x swap classifier, using the new classifier setup, tests for each type of filled order, as well as the necessary logic and helpers needed to support the protocol.


mev_inspect/classifiers/helpers.py
```diff
@@ -84,3 +84,19 @@ def _filter_transfers(
         filtered_transfers.append(transfer)
 
     return filtered_transfers
+
+
+def get_amount_transferred_to_address(address: str, transfers: List[Transfer]) -> int:
+    for transfer in transfers:
+        if transfer.to_address == address:
```
lukevs, line 91: what if there are multiple transfers to the address? also what if they're the same tokens or different

mev_inspect/classifiers/helpers.py
```diff
@@ -84,3 +84,19 @@ def _filter_transfers(
...
+            return transfer.amount
+    return 0
+
+
+def get_amount_transferred_by_token_address(
+    token_address: str, transfers: List[Transfer]
+) -> int:
+    for transfer in transfers:
+        if transfer.token_address == token_address:
```
lukevs, line 100: same question - what if there are multiple

mev_inspect/classifiers/specs/zero_ex.py
```diff
@@ -1,9 +1,78 @@
...
+class ZeroExSwapClassifier(SwapClassifier):
+    @staticmethod
+    def parse_swap(
+        trace: DecodedCallTrace,
+        prior_transfers: List[Transfer],
+        child_transfers: List[Transfer],
+    ) -> Optional[Swap]:
+
+        token_in_amount: int = 0
+        taker_address: str = ""
```
lukevs, line 27: use none instead of empty string for absence

then a string is always a valid value v. sometimes valid

mev_inspect/classifiers/specs/zero_ex.py
```diff
@@ -1,9 +1,78 @@
...
+
+        order: List = trace.inputs["order"]
+
+        token_in_address: str = order[0]
+        token_out_address: str = order[1]
```
lukevs, line 32: is this always exactly 2? silent bug if it's somehow > 2

mev_inspect/classifiers/specs/zero_ex.py
```diff
@@ -1,9 +1,78 @@
...
+
+        if "Rfq" in trace.function_signature:
+
+            taker_address = order[5]
+
+        elif "Limit" in trace.function_signature:
+
+            taker_address = order[6]
```
lukevs, line 40: would be nice to do this more strictly

```
if "Rfq" in trace.function:
    if len(order) != <TODO>:
        raise RuntimeError("...")
```

mev_inspect/classifiers/specs/zero_ex.py
```diff
@@ -1,9 +1,78 @@
...
+
+        else:
+
+            raise RuntimeError(
+                f"Unknown 0x orderbook function: {trace.function_signature}"
+            )
+
+        token_out_amount = trace.inputs["takerTokenFillAmount"]
+
+        if taker_address == "0x0000000000000000000000000000000000000000":
```
lukevs, line 50: should be a constant somewhere

mev_inspect/classifiers/specs/zero_ex.py
```diff
@@ -1,9 +1,78 @@
...
+
+        token_in_amount: int = 0
+        taker_address: str = ""
+
+        order: List = trace.inputs["order"]
+
+        token_in_address: str = order[0]
+        token_out_address: str = order[1]
+
+        if "Rfq" in trace.function_signature:
```
lukevs, line 34: I think better to be explicit and use the full signature. If 0x adds a new `Rfq` function with different ordering, this silently breaks

ex:
```
RFQ_SIGNATURES = [
"fillRfqOrder((address,address,uint128,uint128,address,address,address,bytes32,uint64,uint256),(uint8,uint8,bytes32,bytes32),uint128)",
    ....
]

if trace.function_signature in RFQ_SIGNATURES:
    ...
```

lukevs: also functional kind of does the check on the len of order because you know the signature has that many parameters

right now we're assuming the len of order is valid if Rqf is in

tests/test_0x.py
```diff
@@ -0,0 +1,129 @@
...
+        protocol=Protocol.zero_ex,
+        error=None,
+    )
+
+    block = load_test_block(block_number)
+    trace_classifier = TraceClassifier()
+    classified_traces = trace_classifier.classify(block.traces)
+    result = get_swaps(classified_traces)
+
+    assert swap in result
```
lukevs, line 36: how many times is swap in result? should assert that it shows up once

tests/test_0x.py
```diff
@@ -0,0 +1,129 @@
...
```
lukevs, line 67: same as above

tests/test_0x.py
```diff
@@ -0,0 +1,129 @@
...
```
lukevs, line 98: same as above

tests/test_0x.py
```diff
@@ -0,0 +1,129 @@
...
```
lukevs, line 129: same

mev_inspect/classifiers/helpers.py
lukevs: if we expect there to be exactly one make that assumption explicit not implicit

tests/test_0x.py
lukevs: also maybe assert that it's the only zero x protocol response (aren't multiple for different signatures)

lukevs (commented):
comments

mev_inspect/classifiers/specs/zero_ex.py
```diff
@@ -1,9 +1,97 @@
...
+        ANY_TAKER = "0x0000000000000000000000000000000000000000"
+        RFQ_SIGNATURES = [
+            "fillRfqOrder((address,address,uint128,uint128,address,address,address,bytes32,uint64,uint256),(uint8,uint8,bytes32,bytes32),uint128)",
+            "_fillRfqOrder((address,address,uint128,uint128,address,address,address,bytes32,uint64,uint256),(uint8,uint8,bytes32,bytes32),uint128,address,bool,address)",
+        ]
+        LIMIT_SIGNATURES = [
+            "fillOrKillLimitOrder((address,address,uint128,uint128,uint128,address,address,address,address,bytes32,uint64,uint256),(uint8,uint8,bytes32,bytes32),uint128)",
+            "fillLimitOrder((address,address,uint128,uint128,uint128,address,address,address,address,bytes32,uint64,uint256),(uint8,uint8,bytes32,bytes32),uint128)",
+            "_fillLimitOrder((address,address,uint128,uint128,uint128,address,address,address,address,bytes32,uint64,uint256),(uint8,uint8,bytes32,bytes32),uint128,address,address)",
+        ]
```
lukevs, line 35: constants go up top

mev_inspect/classifiers/specs/zero_ex.py
```diff
@@ -1,9 +1,97 @@
...
+
+class ZeroExSwapClassifier(SwapClassifier):
+    @staticmethod
+    def parse_swap(
+        trace: DecodedCallTrace,
+        prior_transfers: List[Transfer],
+        child_transfers: List[Transfer],
+    ) -> Optional[Swap]:
+
+        ANY_TAKER = "0x0000000000000000000000000000000000000000"
```
lukevs, line 26: this isn't really an accurate name

can just call the `ZERO_ADDRESS`

most ideal would be for the code to read as the explanation

```
def is_any_taker(address: str) -> bool:
    return address == ZERO_ADDRESS
```

lukevs, line 26: `ANY_TAKER_ADDRESS`

mev_inspect/classifiers/helpers.py
```diff
@@ -84,3 +96,69 @@ def _filter_transfers(
...
+            f"A settled order should consist of 2 child transfers, not {len(child_transfers)}."
+        )
+
+    # 2. The function signature must be in the lists of supported signatures
+    if trace.function_signature not in (LIMIT_SIGNATURES + RFQ_SIGNATURES):
+        raise RuntimeError(
+            f"0x orderbook function {trace.function_signature} is not supported"
+        )
+
+    # 3. The swap should not be a child of previous swaps in the block
```
lukevs, line 118: "comments are immediately out of sync with the rest of the codebase"

premium example

mev_inspect/classifiers/helpers.py
```diff
@@ -84,3 +96,69 @@ def _filter_transfers(
...
+
+
+def get_0x_token_in_data(
+    trace: DecodedCallTrace, child_transfers: List[Transfer]
+) -> Tuple[str, int]:
+
+    order: List = trace.inputs["order"]
+    token_in_address = order[0]
+
+    if trace.function_signature in RFQ_SIGNATURES:
```
lukevs, line 145: right now you're checking if `trace.function_signature` is in `RFQ_SIGNATURES` or `LIMIT_SIGNATURES` in another function and then _assuming_ here that whoever calls this function also remembered to call that one

Better to throw in the "else" clause here (where you're using that assumption) than have the check elsewhere

mev_inspect/classifiers/helpers.py
```diff
@@ -84,3 +96,69 @@ def _filter_transfers(
...
+
+
+def _get_taker_token_in_amount(
+    taker_address: str, token_in_address: str, child_transfers: List[Transfer]
+) -> int:
+
+    if taker_address == ANY_TAKER_ADDRESS:
+        for transfer in child_transfers:
+            if transfer.token_address == token_in_address:
+                return transfer.amount
```
lukevs, line 130: same comment as the other function

this function assumes that someone always remembers to call the `is_valid` check to make this function work correctly

better to either have the check in this function or change the data passed into the function to reflect the assumption (ex: `_get_taker_token_in_amount(..., first_transfer, second_transfer)`)

mev_inspect/classifiers/helpers.py
```diff
@@ -84,3 +96,69 @@ def _filter_transfers(
...
+
+    # 2. The function signature must be in the lists of supported signatures
+    if trace.function_signature not in (LIMIT_SIGNATURES + RFQ_SIGNATURES):
+        raise RuntimeError(
+            f"0x orderbook function {trace.function_signature} is not supported"
+        )
+
+    # 3. The swap should not be a child of previous swaps in the block
+
+    return True
```
lukevs, line 120: this function never returns false, it just exists to throw, so better to make it not return anything to make that clear 

ex: `validate_0x_swap`

mev_inspect/classifiers/specs/zero_ex.py
```diff
@@ -1,9 +1,51 @@
...
+        trace: DecodedCallTrace,
+        prior_transfers: List[Transfer],
+        child_transfers: List[Transfer],
+    ) -> Optional[Swap]:
+
+        assert is_valid_0x_swap(trace, child_transfers)
+
+        token_in_address, token_in_amount = get_0x_token_in_data(trace, child_transfers)
+
+        token_out_address, token_out_amount = get_0x_token_out_data(trace)
```
lukevs, line 31: love the breakdown into helper functions ❤️ 

those should live as private functions in here rather than in the "shared" helpers file

mev_inspect/classifiers/specs/zero_ex.py
```diff
@@ -1,9 +1,51 @@
...
+
+class ZeroExSwapClassifier(SwapClassifier):
+    @staticmethod
+    def parse_swap(
+        trace: DecodedCallTrace,
+        prior_transfers: List[Transfer],
+        child_transfers: List[Transfer],
+    ) -> Optional[Swap]:
+
+        assert is_valid_0x_swap(trace, child_transfers)
```
lukevs, line 27: other side of saying this shouldn't return a bool. there's no logical path for the assert to fail

mev_inspect/classifiers/helpers.py
```diff
@@ -84,3 +96,69 @@ def _filter_transfers(
...
+
+    if taker_address == ANY_TAKER_ADDRESS:
+        for transfer in child_transfers:
+            if transfer.token_address == token_in_address:
+                return transfer.amount
+    else:
+        for transfer in child_transfers:
+            if transfer.to_address == taker_address:
+                return transfer.amount
+    return 0
```
lukevs, line 135: is there a case where we ever reach this 0?

mev_inspect/classifiers/helpers.py
gheise: Gotcha

mev_inspect/classifiers/helpers.py
gheise: gotcha

mev_inspect/classifiers/helpers.py
gheise: No, but necessary for mypy to have that return statement

mev_inspect/classifiers/specs/zero_ex.py
```diff
@@ -165,3 +222,57 @@
...
 
 ZEROX_CLASSIFIER_SPECS = ZEROX_CONTRACT_SPECS + ZEROX_GENERIC_SPECS
+
+
+def _get_taker_token_in_amount(
+    taker_address: str, token_in_address: str, child_transfers: List[Transfer]
+) -> int:
+
+    if len(child_transfers) != 2:
+        raise ValueError(
```
lukevs, line 232: Even better than an exception is an interface that can _only be called correctly_

ex:
```
def _get_taker_token_in_amount(
    taker_address: str, token_in_address: str, first_transfer: Transfer, second_transfer: Transfer,
) -> int:
    ...
```

gheise: To get the first and second transfer, i'd still need to check if the length of child_transfers to not get an IndexError, so why not include the exception in the function rather than out of it

lukevs: I think better to include the check outside the function and use the signature to say "here's what I need"

gheise: I disagree, I think keeping the exception inside the function makes more sense. There was an error getting the token_in_data because child_transfers is too long. Keeps it more specific and contextualizes the exception with the use of child_transfers directly below. Makes the necessity for this assumption and its impact clearer in the code

lukevs: Here's better description of why it feels bad to me

If I already have two transfers `first` and `second` this function interface requires that I call the function like
```
_get_taker_token_in_amount(..., [first, second])
```
basically merging it together, to then check a condition which was previously definitely met and now is only maybe met, then breaking it up again

having the exception inside is sort of a "surprise" constraint (i have to read the code to know what's valid) v. making it so i can only call it in a valid way.

gheise: you don't know if you have first and second, until you check child_transfers, which is passed in as an argument. In reality you have:
```
_get_taker_token_in_amount(..., [?])
```

lukevs: why use arguments at all though v. `_get_taker_token_in_amount(args: List[Any])`?

it's the caller's fault that they happen to have a list

if `parse_swaps` was given a dictionary would you change this function to take a dictionary?

if there were two callers where one has a list and one has a dictionary what would be the interface then?

the interface should describe what the function needs, and it shouldn't be possible to call a function with invalid parameters

otherwise i don't know i've passed bad values till a runtime error - instead i could just have that bug not be possible

gheise: it's not. The get_child_traces returns a lit in swaps.py. That list is used for this. But for brevity I rather us close this issue and settle on the solution so we can ship 0x. I don't feel too strongly about this

lukevs (approved):

PR #144, opened 2021-11-24
async based sqlachemy

carlomazzaferro:


listener.py
```diff
@@ -35,16 +33,13 @@ async def run():
...
-
-    inspector = MEVInspector(rpc, inspect_db_session, trace_db_session)
+    inspector = MEVInspector(rpc, None, None)
     base_provider = get_base_provider(rpc)
 
     while not killer.kill_now:
         await inspect_next_block(
             inspector,
-            inspect_db_session,
+            None,
```
lukevs, line 42: need this for `find_latest_block_update` and `update_latest_block`

.gitignore
```diff
@@ -22,3 +22,4 @@ cache
 
 # env
 .envrc
+/out/
```
lukevs, line 25: need this?

mev_inspect/crud/arbitrages.py
```diff
@@ -1,24 +1,18 @@
...
-def delete_arbitrages_for_block(
-    db_session,
-    block_number: int,
-) -> None:
-    (
-        db_session.query(ArbitrageModel)
-        .filter(ArbitrageModel.block_number == block_number)
-        .delete()
+async def delete_arbitrages_for_block(db_session, block_number: int) -> None:
+    await delete_by_block_number(
```
lukevs, line 10: electric - love the helper

mev_inspect/block.py
```diff
@@ -28,13 +29,12 @@ async def create_from_block_number(
...
-    trace_db_session: Optional[orm.Session],
 ) -> Block:
     block: Optional[Block] = None
 
-    if trace_db_session is not None:
-        block = _find_block(trace_db_session, block_number)
-
+    if get_trace_session() is not None:
+        async with get_trace_session() as session:  # type: ignore
+            block = await _find_block(session, block_number)
```
lukevs, line 37: might be what you meant by tracedb issues, but in the inspect-many case we'll open / close a new connection to the DB on each iteration of this loop. would be ideal to share across

.gitignore
carlomazzaferro: For JetBrains-based IDEs, it auto-generates a "compiled" output -- I don't think it hurts to have it

listener.py
carlomazzaferro: Agh, good catch! Forgot to fix this one -- will take a stab at it

mev_inspect/block.py
carlomazzaferro: Updated to have a db session created once per task, similarly to how you structured your PR. Warnings still persist, but it is indeed a nicer set up. Let me know what you think

lukevs: It's better, but I think still better to have it at the even higher level

SQLAlchmey recommends having a single session shared across the duration of the script run v. creating at the task level - ​https://docs.sqlalchemy.org/en/14/orm/session_basics.html#when-do-i-construct-a-session-when-do-i-commit-it-and-when-do-i-close-it

Right now if I call two Inspector functions, it will create a separate session for each

Since we're using the NullPool especially, I believe that means we'll experience the overhead of connecting to the database / closing connection on each call v. getting to reuse the connection across the script.

carlomazzaferro: Totally fair -- I've updated to follow the suggestion. The warnings still show up, but everything works -- please feel free to test it locally! It also seems to be right pattern according to the docs examples regarding usage of async sqlachemy with asyncio.gather: https://docs.sqlalchemy.org/en/14/_modules/examples/asyncio/gather_orm_statements.html

> Note that this pattern loses all transactional safety and is also not
necessarily any more performant than using a single Session, as it adds
significant CPU-bound work both to maintain more database connections
and sessions, as well as within the merging of results from external sessions
into one.

lukevs: Wooo awesome! I'll give a try today!

Wild that they're like "here's how you do it - btw this loses all transactional safety and isn't better"

Ideally we want literally the same thing as the non-async version only without breaking lol. Single threaded and blocking.

lukevs: I wonder if we'll have to switch off the NullPool.

Pretty sure NullPool means exactly one connection at a time. It sounds like from this they create a separate connection for each async usage, which with NullPool might mean opening and closing the connection each time we jump between async functions.

listener.py
```diff
@@ -35,16 +34,15 @@ async def run():
 
     killer = GracefulKiller()
 
-    inspect_db_session = get_inspect_session()
-    trace_db_session = get_trace_session()
-
-    inspector = MEVInspector(rpc, inspect_db_session, trace_db_session)
+    inspector = MEVInspector(rpc)
     base_provider = get_base_provider(rpc)
+    trace_session, inspect_session = get_sessions()
```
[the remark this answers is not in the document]
carlomazzaferro: good catch ser

cli.py
```diff
@@ -67,31 +67,30 @@ async def inspect_many_blocks_command(
...
+        inspect_session=inspect_session,
+        trace_session=trace_session,
     )
 
 
 @cli.command()
 @coro
 async def fetch_all_prices():
-    inspect_db_session = get_inspect_session()
+    inspect_session, _ = get_sessions()
```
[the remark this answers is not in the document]
carlomazzaferro: Good point, otherwise the option would be making these return session makers instead of actual sessions, and refactor the code downstream

mev_inspect/block.py
carlomazzaferro: From your comment it seems to me that actually, maybe the first iteration here should use the pattern of just running everything in a transaction block in sync mode with

```
 await session.run_sync(run_queries)
```

See [example](https://docs.sqlalchemy.org/en/14/_modules/examples/asyncio/greenlet_orm.html)

I'll wait for your feedback on current usage, but I'll do experimentation with this. 

An your notes, I believe you're correct. Each async task would indeed use a new connection, so we might want to switch out of NullPull -- will investigate what are the possibilities there

lukevs: > The function will be passed a synchronous-style Session object and the function can use traditional ORM patterns.

Wow that's pretty sweet - would mean we don't need to change the current crud logic right?

cli.py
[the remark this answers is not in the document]
lukevs: I think make the session instead of returning sessionmaker is kinda nice. I'd imagine we'd just call the sessionmaker immediately most of the time otherwise.

mev_inspect/block.py
lukevs: On this https://docs.sqlalchemy.org/en/14/orm/extensions/asyncio.html#using-multiple-asyncio-event-loops
it says
> An application that makes use of multiple event loops, for example by combining asyncio with multithreading, should not share the same AsyncEngine with different event loops when using the default pool implementation.
which makes me think it does actually re-use the same connections across async

I'll run a test on this branch and report back

lukevs: Trying locally, I did see it create multiple connections. And then it seemed like everything locked up.

To replicate, I did
```
./mev inspect-many 13115099 13115299
```

Then was checking DB connections by running
```
select pid as process_id,
       usename as username,
       datname as database_name,
       client_addr as client_address,
       application_name,
       backend_start,
       state,
       state_change
from pg_stat_activity;
```

which looked like this for a bit
```
 process_id | username | client_address | application_name |        state
------------+----------+----------------+------------------+---------------------
        997 |          |                |                  |
        999 | postgres |                |                  |
       1399 | postgres | 10.244.0.55    | psql             | active
       2523 | postgres | 10.244.0.59    |                  | idle
       2525 | postgres | 10.244.0.59    |                  | idle in transaction
       2527 | postgres | 10.244.0.59    |                  | idle in transaction
       2529 | postgres | 10.244.0.59    |                  | active
       2531 | postgres | 10.244.0.59    |                  | idle in transaction
```

then froze like this
```
 process_id | username | client_address | application_name | state
------------+----------+----------------+------------------+--------
        997 |          |                |                  |
        999 | postgres |                |                  |
       1399 | postgres | 10.244.0.55    | psql             | active
       2523 | postgres | 10.244.0.59    |                  | active
       2525 | postgres | 10.244.0.59    |                  | active
       2527 | postgres | 10.244.0.59    |                  | active
       2529 | postgres | 10.244.0.59    |                  | active
       2531 | postgres | 10.244.0.59    |                  | active
```

lukevs:
Talked over discord - we're going to hold on this change for now

PR #146, opened 2021-11-26
Convert block_timestmap from numeric to timestamp

lukevs:
More explicit and prevents us from doing a cast when comparing with `prices` table


bertmiller (approved):
lgtm

Issue #145, opened 2021-11-26
mev-inspect fails to inspect block 13486138


lukevs:
Block 13486138 contains some swaps with a protocol BankETH
https://etherscan.io/tx/0x08fdd7cb5fb68cb41fb740362c3f98116d37c6b1234314b81a394fcc81a90099

It's getting parsed as an arbitrage with the same swap happening twice which then fails to write to the DB

## Replicate
```
./mev inspect 13486138
```

gheise:
This also occurs on block 13750068, transaction 0xcd360c9085191c24c5346226dbbf04f5d93e00be616b972c1c560747fa7eb954

PR #153, opened 2021-12-06
Fix arbitrage swap double entry bug

gheise:
## Resolution to Issue #145 
As highlighted in the issue, there was a bug in getting arbitrages where, in certain cases, a start/end pair with equal entries would be passed as a valid. This causes the arbitrage swaps list to contain a double entry, which breaks when writing to the db due to a non-unique primary key. The error occurred every couple hundred blocks. A check can be introduced to remove this from the valid start/end pairs.

## TLDR
- Introduce a check for valid arbitrage start/end pairs where the start/end swaps cannot be the same swap

### DIFF
--- a/mev_inspect/arbitrages.py
+++ b/mev_inspect/arbitrages.py
@@ -88,14 +88,14 @@ def _get_all_start_end_swaps(swaps: List[Swap]) -> List[Tuple[Swap, Swap]]:
     """
     pool_addrs = [swap.contract_address for swap in swaps]
     valid_start_ends: List[Tuple[Swap, Swap]] = []
-    for potential_start_swap in swaps:
-        for potential_end_swap in swaps:
+    for index, potential_start_swap in enumerate(swaps):
+        remaining_swaps = swaps[:index] + swaps[index + 1 :]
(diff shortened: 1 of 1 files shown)


mev_inspect/arbitrages.py
```diff
@@ -95,6 +95,7 @@ def _get_all_start_end_swaps(swaps: List[Swap]) -> List[Tuple[Swap, Swap]]:
                 == potential_end_swap.token_out_address
                 and potential_start_swap.from_address == potential_end_swap.to_address
                 and not potential_start_swap.from_address in pool_addrs
+                and potential_start_swap != potential_end_swap
```
lukevs, line 98: rather than comparing the objects directly which can sometimes be funky, I think it'd be slightly better to explicitly say what you mean which is "skip the swap I have as the start swap"

something like:
```
for index, potential_start_swap in enumerate(swaps):
    other_swaps = swaps[:index] + swaps[index+1:]
    for potential_end_swap in other_swaps:
        ...
```

gheise: Thanks for the feedback, this was implemented

lukevs (approved):

PR #160, opened 2021-12-14
Write transaction position for swaps and traces

lukevs:
Needed for sandwiches

Merge after https://github.com/flashbots/mev-inspect-py/pull/161 is applied

### DIFF
--- a/mev_inspect/models/swaps.py
+++ b/mev_inspect/models/swaps.py
@@ -8,6 +8,7 @@ class SwapModel(Base):
 
     abi_name = Column(String, nullable=False)
     transaction_hash = Column(String, primary_key=True)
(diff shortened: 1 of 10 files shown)


mev_inspect/models/swaps.py
```diff
@@ -8,6 +8,7 @@ class SwapModel(Base):
 
     abi_name = Column(String, nullable=False)
     transaction_hash = Column(String, primary_key=True)
+    transaction_position = Column(Numeric, nullable=True)  # TODO - backfill
```
gheise, line 11: I would remove this comment, odds of being forgotten here are high

gheise (approved):

PR #161, opened 2021-12-14
Add nullable transaction position field

lukevs:
Needed for sandwiches

PR #162, opened 2021-12-14
Add sandwiches tables

lukevs:
Adds a table for sandwiches and another for the swaps being sandwiched

Currently doesn't include profit data as this table's immediate purpose is for filtering arbitrages that participate in sandwiches


bertmiller:
Oh I see now, #161 builds on #162.

lukevs:
@bertmiller my bad - I must have flopped the ordering

PR #163, opened 2021-12-14
Add sandwiches

lukevs:
Adds support for finding and writing sandwiches to the DB

The logic is to:
- Get all swaps
- Order them by transaction position - the order they're processed in the block
- For each swap (frontrun candidate), try to find a sandwich that it starts - this is defined as:
  - At least one swap of the same pool and tokens in the same direction as the frontrun candidate swap
  - A swap in the opposite direction as the frontrun candidate swap of the same tokens and initiated by the same address


mev_inspect/models/sandwiches.py
```diff
@@ -0,0 +1,16 @@
...
+    __tablename__ = "sandwiches"
+
+    id = Column(String, primary_key=True)
+    created_at = Column(TIMESTAMP, server_default=func.now())
+    block_number = Column(Numeric, nullable=False)
+    sandwicher_address = Column(String(256), nullable=False)
+    frontrun_swap_transaction_hash = Column(String(256), nullable=False)
+    frontrun_swap_trace_address = Column(ARRAY(Integer), nullable=False)
+    backrun_swap_transaction_hash = Column(String(256), nullable=False)
+    backrun_swap_trace_address = Column(ARRAY(Integer), nullable=False)
```
gheise, line 16: Why are the victim transaction hashes / trace addresses not written to the database?

gheise:
conflict with main

mev_inspect/models/sandwiches.py
lukevs: they are - they're in the join table

lukevs: https://github.com/flashbots/mev-inspect-py/pull/163/files#diff-f46030029e3258add72e9afe20c5f355d2d30d93c380505b22ea52966a6e563cR54

gheise: Is there a specific reason to separate the victim transaction data from the bot transaction data in the database? It seems simpler to include both under the sandwiches table, rather than splitting it into two tables, and therefore removing the join on sandwich id

lukevs: because there can be multiple victims but there's only one bot

PR #164, opened 2021-12-14
Fix 0x reverted orders bug 

gheise:
## Overview
Some 0x orders are reverted which fails in the swap classifier during the search for the transfers.

>> https://etherscan.io/tx/0xdc6259a4bc95044e23290f6b6e894e0ae7aa8474a3e38218b272e47a50ad297e

 Introduces a check for the status of the order and only throws the error in the case the transaction is not reverted.

### DIFF
--- a/mev_inspect/classifiers/specs/zero_ex.py
+++ b/mev_inspect/classifiers/specs/zero_ex.py
@@ -231,7 +231,10 @@ def _get_taker_token_in_amount(
     child_transfers: List[Transfer],
 ) -> int:
 
-    if len(child_transfers) != 2 and trace.error == None:
+    if trace.error is not None:
+        return 0
+
+    if len(child_transfers) != 2:
         raise ValueError(
             f"A settled order should consist of 2 child transfers, not {len(child_transfers)}."
         )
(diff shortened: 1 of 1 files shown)


mev_inspect/classifiers/specs/zero_ex.py
```diff
@@ -225,10 +225,13 @@ def parse_swap(
...
 def _get_taker_token_in_amount(
-    taker_address: str, token_in_address: str, child_transfers: List[Transfer]
+    trace: DecodedCallTrace,
+    taker_address: str,
+    token_in_address: str,
+    child_transfers: List[Transfer],
 ) -> int:
 
-    if len(child_transfers) != 2:
+    if len(child_transfers) != 2 and trace.error == None:
```
lukevs, line 234: none checks are always `is None` or `is not None`

https://legacy.python.org/dev/peps/pep-0008/#programming-recommendations

lukevs: I think rather than having the “trickle through” down to that zero, it’d be nice to make it more explicit as its own check up front

```
if trace.error is not None:
    return 0
```

i don’t think we have any known cases besides this where that last zero should get hit right? in which case it should become an exception

lukevs (approved):

PR #178, opened 2021-12-21
Bulk delete and write data

lukevs:
## Current state
Currently our backfills spend a lot of time interacting with Postgres

We delete all + write all new data on a per-block basis which requires a certain amount of overhead

We'd like to take advantage of more bulk delete / write strategies

## This PR
As a first step towards optimizing these writes, this changes our logic from writing / deleting per block to allowing us to bulk get all of the data for a chunk of blocks then write it

The parameters as is are that we write every 10 blocks

This also adds optimizations around bulk writing using psycopg2 fast execution helpers

https://docs.sqlalchemy.org/en/13/dialects/postgresql.html#psycopg2-fast-execution-helpers

## Results

Locally, this results in a speedup from 40 to 30 seconds when inspecting 10 blocks

### DIFF
--- a/mev_inspect/crud/arbitrages.py
+++ b/mev_inspect/crud/arbitrages.py
@@ -4,17 +4,20 @@ from uuid import uuid4
 from mev_inspect.models.arbitrages import ArbitrageModel
 from mev_inspect.schemas.arbitrages import Arbitrage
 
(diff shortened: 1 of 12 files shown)


gheise:
Why 10?

gheise (approved):

PR #180, opened 2021-12-21
New error column for arbitrages

taarushv:


mev_inspect/arbitrages.py
```diff
@@ -52,6 +52,11 @@ def _get_arbitrages_from_swaps(swaps: List[Swap]) -> List[Arbitrage]:
         ]
         routes = _get_all_routes(start, end, potential_intermediate_swaps)
 
+        error = ""
```
lukevs, line 55: None instead of empty string

Then if it's any string (is not None) we know it's a valid error

mev_inspect/arbitrages.py
```diff
@@ -52,6 +52,11 @@ def _get_arbitrages_from_swaps(swaps: List[Swap]) -> List[Arbitrage]:
...
+        for swap in swaps:
```
lukevs, line 56: Not all swaps will be involved in all arbs

Needs to be considered separately for each route I think

Like
```
for route in routes:
    error = None

    for swap in route:
        if swap.error is not None
             error = swap.error
 
     ...
     arb = Arbitrage(..., error=error)
```

lukevs (approved):

Issue #176, opened 2021-12-21
Duplicate arbitrages


lukevs:
Transaction '0x565146ec57af69208b4a37e3a138ab85c6a6ff358fffb0077824a7378a67c4d6' creates 19 instances of the same arbitrage

That shouldn't happen

lukevs:
More in block 11421847

PR #183, opened 2021-12-22
Only use each swap in a single arbitrage

lukevs:
Fixes https://github.com/flashbots/mev-inspect-py/issues/176

## Current
We allow a swap to participate in multiple arbs

This causes bad data and also leads to some nasty duplicate issues (like getting stuck creating 13k arbs https://github.com/flashbots/mev-inspect-py/issues/176)

## This PR
Picks the shortest length route as the best route for a swap to participate in

If I do two arbs like:
A =pool 1=> B =pool 2=> A
and
A =pool 3=> C =pool 4=> A

picking the shortest would create 2 arbs instead of one long one involving all, which would mean the same total profit but with a more specific description of what's happening

## Testing

Verified all blocks in https://github.com/flashbots/mev-inspect-py/issues/176 work now

All tests pass

Looking at block `11421847` looking at profit you can see the same tokens being wrapped around multiple times
image

### DIFF
--- a/mev_inspect/arbitrages.py
+++ b/mev_inspect/arbitrages.py
@@ -93,7 +93,6 @@ def _get_shortest_route(
     end_swaps: List[Swap],
     all_swaps: List[Swap],
 ) -> Optional[List[Swap]]:
-    # TODO - add max length
     for end_swap in end_swaps:
         if start_swap.token_out_address == end_swap.token_in_address:
             return [start_swap, end_swap]
--- a/mev_inspect/arbitrages.py
+++ b/mev_inspect/arbitrages.py
@@ -1,5 +1,5 @@
 from itertools import groupby
-from typing import List, Tuple
+from typing import List, Optional, Tuple
 
 from mev_inspect.schemas.arbitrages import Arbitrage
 from mev_inspect.schemas.swaps import Swap
@@ -45,37 +45,14 @@ def _get_arbitrages_from_swaps(swaps: List[Swap]) -> List[Arbitrage]:
     if len(start_ends) == 0:
         return []
 
-    used_swaps = []
+    used_swaps: List[Swap] = []
 
     for (start, ends) in start_ends:
-        if start in used_swaps:
-            continue
+        route = _get_shortest_route_from_start(start, ends, swaps, used_swaps)
 
-        shortest_route = None
-
-        for end in ends:
-            if end in used_swaps:
-                continue
-
-            potential_intermediate_swaps = [
-                swap
-                for swap in swaps
-                if (swap is not start and swap is not end and swap not in used_swaps)
-            ]
-
-            routes = _get_all_routes(
-                start,
-                end,
-                potential_intermediate_swaps,
-            )
-
-            for route in routes:
-                if shortest_route is None or len(route) < len(shortest_route):
-                    shortest_route = route
-
-        if shortest_route is not None:
-            start_amount = shortest_route[0].token_in_amount
-            end_amount = shortest_route[-1].token_out_amount
+        if route is not None:
+            start_amount = route[0].token_in_amount
+            end_amount = route[-1].token_out_amount
             profit_amount = end_amount - start_amount
             error = None
             for swap in route:
@@ -83,11 +60,11 @@ def _get_arbitrages_from_swaps(swaps: List[Swap]) -> List[Arbitrage]:
                     error = swap.error
 
             arb = Arbitrage(
-                swaps=shortest_route,
-                block_number=shortest_route[0].block_number,
-                transaction_hash=shortest_route[0].transaction_hash,
-                account_address=shortest_route[0].from_address,
-                profit_token_address=shortest_route[0].token_in_address,
+                swaps=route,
+                block_number=route[0].block_number,
+                transaction_hash=route[0].transaction_hash,
+                account_address=route[0].from_address,
+                profit_token_address=route[0].token_in_address,
                 start_amount=start_amount,
                 end_amount=end_amount,
                 profit_amount=profit_amount,
@@ -95,7 +72,7 @@ def _get_arbitrages_from_swaps(swaps: List[Swap]) -> List[Arbitrage]:
             )
 
             all_arbitrages.append(arb)
-            used_swaps.extend(shortest_route)
+            used_swaps.extend(route)
 
     if len(all_arbitrages) == 1:
         return all_arbitrages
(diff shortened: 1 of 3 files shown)


mev_inspect/arbitrages.py
```diff
@@ -78,7 +83,49 @@ def _get_arbitrages_from_swaps(swaps: List[Swap]) -> List[Arbitrage]:
...
+            return [start_swap, end_swap]
+
+    other_swaps = [
+        swap for swap in all_swaps if (swap is not start_swap and swap not in end_swaps)
+    ]
+
+    if len(other_swaps) == 0:
+        return None
+
+    shortest_route_rest = None
```
gheise, line 102: what is shortest_route_rest meant to indicate? It's a bit confusing, something along the lines current_shortest_route?

mev_inspect/arbitrages.py
```diff
@@ -78,7 +83,49 @@ def _get_arbitrages_from_swaps(swaps: List[Swap]) -> List[Arbitrage]:
...
+                other_swaps,
+            )
+
+            if shortest_from_next is not None and (
+                shortest_route_rest is None
+                or len(shortest_from_next) < len(shortest_route_rest)
+            ):
+                shortest_route_rest = shortest_from_next
+
+    if shortest_route_rest is None:
```
gheise, line 122: Under what circumstances do we expect shortest_route_rest to be None?

mev_inspect/arbitrages.py
```diff
@@ -78,7 +83,49 @@ def _get_arbitrages_from_swaps(swaps: List[Swap]) -> List[Arbitrage]:
...
+        return None
+    else:
+        return [start_swap] + shortest_route_rest
+
+
+def _get_all_start_end_swaps(swaps: List[Swap]) -> List[Tuple[Swap, List[Swap]]]:
```
gheise, line 128: The return type indicates this will return a list of tuples containing a start swap and a list of corresponding plausible end swaps. The name/docstring seems a bit misleading that it would return 'pairs', as it's really returning all possible ends for each start swap.

tests/test_arbitrages.py
```diff
@@ -171,39 +171,46 @@ def test_three_pool_arbitrage(get_transaction_hashes, get_addresses):
...
 
 
-def test_get_all_routes():
+def test_get_shortest_route():
     # A -> B, B -> A
     start_swap = create_generic_swap("0xa", "0xb")
     end_swap = create_generic_swap("0xb", "0xa")
-    routes = _get_all_routes(start_swap, end_swap, [])
-    assert len(routes) == 1
+    route = _get_shortest_route(start_swap, [end_swap], [])
```
gheise, line 178: ```shortest_route = _get_shortest_route(...)```
sounds more aligned with what the test is doing. The following length check is meant to assert the shortest _route is 2, small comment but would be more legible. Same thing applies for the following cases and the 'route' declared in arbitrages.py

tests/test_arbitrages.py
```diff
@@ -171,39 +171,46 @@ def test_three_pool_arbitrage(get_transaction_hashes, get_addresses):
...
     other_swaps = [create_generic_swap("0xb", "0xa"), create_generic_swap("0xb", "0xc")]
-    routes = _get_all_routes(start_swap, end_swap, other_swaps)
-    assert len(routes) == 1
-    expect_simple_route = [["0xa", "0xb"], ["0xb", "0xc"], ["0xc", "0xa"]]
-    assert len(routes[0]) == len(expect_simple_route)
-    for i in range(len(expect_simple_route)):
-        assert expect_simple_route[i][0] == routes[0][i].token_in_address
-        assert expect_simple_route[i][1] == routes[0][i].token_out_address
+    route = _get_shortest_route(start_swap, [end_swap], other_swaps)
+    expected_smallest_route = [["0xa", "0xb"], ["0xb", "0xc"], ["0xc", "0xa"]]
```
gheise, line 205: shortest

tests/test_arbitrages.py
```diff
@@ -171,39 +171,46 @@ def test_three_pool_arbitrage(get_transaction_hashes, get_addresses):
...
+
+    assert route is not None
+    assert len(route) == len(expected_smallest_route)
```
gheise, line 208: ```len(observed_shortest_route) == len(expected_shortest_route)```
 maybe? More specific

gheise (commented):
Wagmi

mev_inspect/arbitrages.py
lukevs: it’s the shortest path through the rest of the nodes to one of the ends

mev_inspect/arbitrages.py
lukevs: when there is no path to the ends

mev_inspect/arbitrages.py
lukevs: good call - will update

tests/test_arbitrages.py
lukevs: 👀

tests/test_arbitrages.py
lukevs: actual / expected is common

can i switch to that

tests/test_arbitrages.py
lukevs: we’re not specifically checking that it’s the shortest here though. just that a route is found

tests/test_arbitrages.py
lukevs: 👍

mev_inspect/arbitrages.py
lukevs: (comments always out of sync with the codebase 😉)

tests/test_arbitrages.py
gheise: so why is the check on the length and not on existence? A check on the length seems intended to check just that - the length of the route, not only that this route exists but that it is of a certain size

tests/test_arbitrages.py
gheise: That's great

mev_inspect/arbitrages.py
gheise: Got it

mev_inspect/arbitrages.py
gheise: is it not used in this case as a the current of the current/next variable pair to find the shortest route recursively?

lukevs: i should add a type to it

it’s Optional[List[Swap]]

it holds the shortest path so far as we’re iterating through each of the potential next start nodes

high level is like

get shortest(start, ends)
- find all potential continuations
- for each continuation, get shortest path from that one to an end (get_shortest(next, ends))
- return start + shortest of all continuations

tests/test_arbitrages.py
lukevs: it’s just checking it’s the route we expect. probably would be better to check like we do with the others

tests/test_arbitrages.py
gheise: that's great

mev_inspect/arbitrages.py
gheise: shortest path so far -> current_shortest_path

This seems like a more descriptive name

lukevs: it's not current because it isn't the full path

maybe `shortest_next_path`

lukevs: wnet with "shortest_remaining_route"

lukevs:
gui said good on this

Issue #181, opened 2021-12-22
OverflowError: cannot fit 'int' into an index-sized integer


lukevs:
To reproduce:
```
./mev inspect 11101564
```

lukevs:
Etherscan also fails to decode - https://etherscan.io/tx/0x4ed45f738f2fb8a74a0182bc70cd5110605dc931eba3de7bac9626b0ff77fa37

PR #189, opened 2021-12-23
Ignore overflow errors on trace decode

lukevs:
If it's overflowing the expected signature, it's no good

Etherscan also can't decode https://etherscan.io/tx/0xed361a0c8cb8a287c0faacaa5f33bdea725b35d9692a21f7a46eb00dee26aaada

Solves this issue:
https://github.com/flashbots/mev-inspect-py/issues/181


lukevs:
Gui said this is good

PR #195, opened 2021-12-25
Use middleware for all RPC calls

lukevs:
## Right now

We call eth_getBlockReceipts and trace_block directly from the base_provider
```
base_provider.make_request("eth_getBlockReceipts", [block_number]),
base_provider.make_request("trace_block", [block_number]),
```

Doing that bypasses the middleware in w3, most importantly the retry middleware

That means if either of these methods fail, they won't be retried

## This PR

Adds support for both of these methods on AsyncEth
Adds them to a list of retry-able methods
Calls them using w3

This has the bonus benefit of removing the need for base provider to be passed around

### DIFF
--- a/mev_inspect/block.py
+++ b/mev_inspect/block.py
@@ -24,7 +24,6 @@ async def get_latest_block_number(base_provider) -> int:
 
 
 async def create_from_block_number(
-    base_provider,
     w3: Web3,
     block_number: int,
     trace_db_session: Optional[orm.Session],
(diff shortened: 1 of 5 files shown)


mev_inspect/block.py
```diff
@@ -35,34 +34,22 @@ async def create_from_block_number(
...
-        logger.warning(
-            f"Failed to create objects from block: {block_number}: {e}, retrying: {retries + 1} / 3"
-        )
-        if retries < 3:
-            await asyncio.sleep(5)
-            return await _fetch_block(w3, base_provider, block_number, retries)
-        else:
-            raise
+    receipts: List[Receipt] = [Receipt(**receipt) for receipt in receipts_json]
+    traces = [Trace(**trace_json) for trace_json in traces_json]
```
lukevs, line 52: "result" is selected automatically - and our middleware handles retries so this doesn't need to

Issue #198, opened 2021-12-29
No support for sandwiches through multiple pools


lukevs:
## Description

A bot wants to sandwich a large trade on Pool 1 going from token A to token B

The bot currently holds token B, but not A

To frontrun this transaction, the bot needs to go A => B on Pool 1

So it will
- Trade B => A on some Pool 2
- Trade A => B on Pool 1

Then do the reverse for the backrun

Currently we don't pick this up because we filter for tokens to be from the sandwicher address

## Example

Block 13894234 should include a sandwich

first transaction in the block trades WETH => SUSHI on ShibaSwap (https://etherscan.io/address/0x6cbefa95e42960e579c2a3058c05c6a08e2498e9) then SUSHI => WETH on SushiSwap
https://etherscan.io/tx/0xada77bc134f33628c0f7e1698db5ef13a1d5f302c18a90e80aca3a668e042cba

then sandwiched transaction trades SUSHI => WETH on Sushi
https://etherscan.io/tx/0xb2f70d8400e39797dd0dbb1b5883fb76d4a8804579b2d15425cf6ec4db3d5077

then the same bot does the first tx in reverse (WETH => SUSHI on Sushi, SUSHI => WETH on Shiba)
https://etherscan.io/tx/0x3d19c140aeca2076acdb42873569b49e3dd0764b01f48a7831341dc2e97e54b4

lukevs:
This is fixed now
Verified this is correctly tagged

PR #199, opened 2021-12-29
Support sandwiches including multiple pools

lukevs:
Fixes this: https://github.com/flashbots/mev-inspect-py/issues/198

For sandwiches, the backrun swap doesn't need to be funded by the sandwicher, it just should end up cycling back to their account

### DIFF
--- a/mev_inspect/sandwiches.py
+++ b/mev_inspect/sandwiches.py
@@ -46,7 +46,7 @@ def _get_sandwich_starting_with_swap(
             elif (
                 other_swap.token_out_address == front_swap.token_in_address
                 and other_swap.token_in_address == front_swap.token_out_address
-                and other_swap.from_address == sandwicher_address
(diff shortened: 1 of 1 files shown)


bertmiller:
` if (
                other_swap.token_in_address == front_swap.token_in_address
                and other_swap.token_out_address == front_swap.token_out_address
                and other_swap.from_address != sandwicher_address
            ):
                sandwiched_swaps.append(other_swap)`
Does this work if it's a sandwich of a Token -> ETH trade and the sandwich needs to go from ETH-> token -> ETH?

In that case the token in of the first swap would be eth, even though the sandwiched trades would be token.

lukevs:
It'll still work because that ETH => Token => ETH cycle will break into two swaps, and we'll say the "front swap" is the Token => ETH piece

lukevs:
@bertmiller said lgtm

Issue #155, opened 2021-12-10
cannot run backfill


simple4016:
i can run all the others except the backfill.
when run backfill:
<img width="1917" alt="CleanShot 2021-12-10 at 11 44 54@2x" src="https://user-images.githubusercontent.com/5715598/145513831-6339f1e5-34a1-4dc1-8620-e730514014c6.png">

lukevs:
Hmm it looks like the pod is maybe loading a pre-python 3.5 version

Are you using the full kubernetes / tilt setup?

simple4016:
> Hmm it looks like the pod is maybe loading a pre-python 3.5 version
> 
> Are you using the full kubernetes / tilt setup?

thanks for your reply.

i go step by step in the Readme.

lukevs:
If you run
```
./mev exec python --version
```
what do you see?

simple4016:
Python 3.9.9

lukevs:
We were using the default python instead of poetry installed python for the backfill script

This PR fixes the issue: https://github.com/flashbots/mev-inspect-py/pull/200

lukevs:
Fixed - Thanks @simple4016 for reporting!

PR #200, opened 2021-12-30
Use poetry for backfill script

lukevs:
Fixes: https://github.com/flashbots/mev-inspect-py/issues/155

Otherwise we're using the system default python


bertmiller (approved):
LGTM

PR #202, opened 2021-12-31
Queue backfills with Redis

lukevs:
## Right now

To backfill a large set of blocks, we:
- Specify the range to backfill and a parallelism N
- Spin up N pods each dedicated to len(range) / N blocks
- Wait for each pod to complete

Some issues with this:
- It's difficult to tell how quickly we're processing the block range / how close we are to done
- If one block fails in the range, that entire pod fails causing it to start over
- if pods finish early, the parallelism drops (only the remaining ones keep working) - we'd prefer to have parallelism of N throughout the full time
- If we want to increase or decrease parallelism during the run, we're out of luck
- If we want to add more blocks to this backfill without changing parallelism, we're out of luck

## This PR

Switches to a new model using a simple queue over Redis called dramatiq

Now to backfill a range we:
- Queue up many small batches (ex: 10 blocks) of work for a large range
- Spin up a set parallelism of worker nodes to process one batch at a time
- If a batch fails, it's automatically retried
- if it continues to fail after retries, it stops being retried and is placed in a "dead letter" queue for manual inspection, allowing the pod to continue working on the rest of the data

Using this model we can:
- Query redis to see how far along we are and any blocks being retried or failed
- Increase or decrease our parallelism by adjusting the number of workers
- Increase or decrease the number of blocks being worked on by adding / removing elements from Redis

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,13 @@ You can see worker pods spin up then complete by watching the status of all pods
 watch kubectl get pods
 ```
 
+To see progress and failed batches, connect to Redis with
+```
+./mev redis
+```
+
+then query keys keys and values using the [spec shared by dramatiq](https://github.com/Bogdanp/dramatiq/blob/24cbc0dc551797783f41b08ea461e1b5d23a4058/dramatiq/brokers/redis/dispatch.lua#L24-L43)
+
 To watch the logs for a given pod, take its pod name using the above, then run:
 ```
 kubectl logs -f pod/mev-inspect-backfill-abcdefg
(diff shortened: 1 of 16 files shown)


bertmiller:
The high level idea makes sense to me, pinging @metachris for a review

metachris (approved):

metachris:
Makes sense, and sounds like a good approach! Not too familiar with the code, but the changes look straight forward. If those are battle-tested, I think this PR is good 👍

PR #210, opened 2022-01-03
Improve dockerfile

tmikulin:
1.) Changed the parent docker image to buster slim (still debian based but a lot more slimmer)
2.) Added a separate flashbot user (for security), it doesn't run as the root user.
3.) Also worth mentioning, since the container now runs as a non root user you won't be able to install any stuff in the image.

The resulting docker image is half the size of the original one, which will come in handy with the CI/CD stuff.
I'm not sure how you handle the CI/CD part, I didn't find any build info (I guess for now you use git pull or something like that)

Tom


Dockerfile
```diff
@@ -1,22 +1,29 @@
...
+RUN poetry config virtualenvs.create false \
+    && poetry install
 
-COPY . /app
+COPY --chown=flashbot . /app
 
-# easter eggs 😝
-RUN echo "PS1='🕵️:\[\033[1;36m\]\h \[\033[1;34m\]\W\[\033[0;35m\]\[\033[1;36m\]$ \[\033[0m\]'" >> ~/.bashrc
+# easter eggs 
+RUN echo "PS1='️ :\[\033[1;36m\]\h \[\033[1;34m\]\W\[\033[0;35m\]\[\033[1;36m\]$ \[\033[0m\]'" >> ~/.bashrc
```
lukevs, line 26: Lost the emojis!

lukevs:
Did some tests locally and all looks good!

Only comment is those emojis :) 

Then good to go!

Dockerfile
tmikulin: added the emojis!

lukevs:
Merged - thanks @tmikulin !!

tmikulin:
np @lukevs

PR #212, opened 2022-01-03
Cut out early from arbitrages if we've already found a shorter path

lukevs:
If multiple swaps can form a circuit (arbitrage) in multiple ways, we create an arbitrage from the shortest circuit.

If multiple paths are equally short, we pick the first one.

Currently, as we're looking for shortest routes, if we find a route with length ex: 4, we'll still look for "shorter" candidate routes with depth 4, 5, 6 etc even though they'll be rejected once they're ultimately returned.

This PR adds a "shortest_route_length" parameter to communicate downward the current shortest length so we can cut out early if there are no ways to form shorter routes

### DIFF
--- a/mev_inspect/arbitrages.py
+++ b/mev_inspect/arbitrages.py
@@ -92,10 +92,20 @@ def _get_shortest_route(
     start_swap: Swap,
     end_swaps: List[Swap],
     all_swaps: List[Swap],
+    shortest_route_length: Optional[int] = None,
 ) -> Optional[List[Swap]]:
+    if len(end_swaps) == 0:
+        return None
+
+    if shortest_route_length is not None and shortest_route_length <= 2:
+        return None
+
     for end_swap in end_swaps:
         if start_swap.token_out_address == end_swap.token_in_address:
             return [start_swap, end_swap]
+
+    if shortest_route_length is not None and shortest_route_length == 3:
+        return None
 
     other_swaps = [
         swap for swap in all_swaps if (swap is not start_swap and swap not in end_swaps)
@@ -105,6 +115,9 @@ def _get_shortest_route(
         return None
 
     shortest_remaining_route = None
+    shortest_remaining_route_length = (
+        None if shortest_route_length is None else shortest_route_length - 1
+    )
 
     for next_swap in other_swaps:
         if start_swap.token_out_address == next_swap.token_in_address and (
@@ -116,6 +129,7 @@ def _get_shortest_route(
                 next_swap,
                 end_swaps,
                 other_swaps,
+                shortest_route_length=shortest_remaining_route_length,
             )
 
             if shortest_from_next is not None and (
@@ -123,6 +137,7 @@ def _get_shortest_route(
                 or len(shortest_from_next) < len(shortest_remaining_route)
             ):
                 shortest_remaining_route = shortest_from_next
+                shortest_remaining_route_length = len(shortest_from_next)
 
     if shortest_remaining_route is None:
         return None


mev_inspect/arbitrages.py
```diff
@@ -92,11 +92,21 @@ def _get_shortest_route(
...
+        return None
+
+    if max_route_length is not None and max_route_length < 2:
+        return None
+
     for end_swap in end_swaps:
         if start_swap.token_out_address == end_swap.token_in_address:
             return [start_swap, end_swap]
 
+    if max_route_length is not None and max_route_length == 2:
```
gheise, line 107: Is there a reason to separate this line from:

``` if max_route_length is not None and max_route_length < 2:```

why not:

``` if max_route_length is not None and max_route_length <= 2``` ?

File: mev_inspect/arbitrages.py [hunk 2]
```diff
@@ -105,6 +115,9 @@ def _get_shortest_route(
         return None
 
     shortest_remaining_route = None
+    max_remaining_route_length = (
```
gheise, line 118: ```max_remaining_route_length``` implies that the variable represents the longest possible route that is still viable - which makes sense - besides the fact that initial assignment is ```None``` - which is meaningless. What about ```current_shortest_route_length```? Then, ```None``` indicates there is no current shortest route, until the function is called at runtime

mev_inspect/arbitrages.py
```diff
@@ -116,13 +129,15 @@ def _get_shortest_route(
                 next_swap,
                 end_swaps,
                 other_swaps,
+                max_route_length=max_remaining_route_length,
```
gheise, line 132: The earlier suggestions would make this ```shortest_route_length = current_shortest_route_length```, much more representative of what's going on: only search for routes below the upper limit: the current shortest route

mev_inspect/arbitrages.py
gheise, line 107: When do we expect this to be 2?

mev_inspect/arbitrages.py
```diff
@@ -92,11 +92,21 @@ def _get_shortest_route(
     start_swap: Swap,
     end_swaps: List[Swap],
     all_swaps: List[Swap],
+    max_route_length: Optional[int] = None,
```
gheise, line 95: Below I left some reasons why I think this should be ```shortest_route_length```

gheise (commented):
Clever :+1:

mev_inspect/arbitrages.py
lukevs: because if it were less than 2 it'd have already cut out on line 100

lukevs: when the shortest path is length 3

mev_inspect/arbitrages.py
lukevs: `current_shortest_route_length` isn't accurate because it isn't the shortest route. It's the shortest of the rest of the route (shortest continuation) - so it'd be:
`current_shortest_remaining_route_length`

but still max and shortest are slightly distinct

i moved away from shortest because it ends up getting messy looking in the logic

ex: if the current shortest path is 4, we should only be looking for paths that are max length 3

so you have to change all the comparisons to be "one less than the shortest length"

v. max says "Optionally specify a max length for a shortest path"

mev_inspect/arbitrages.py
gheise: Right, that's my point; is there a reason to not include them in the same line?

mev_inspect/arbitrages.py
gheise: what about ```next_shortest_route_length```?

mev_inspect/arbitrages.py
gheise: Why should that return None?

lukevs: yes because the line above it can return the case of len 2, but nowhere below can

lukevs: the parameter itself says "get the shortest route with a max length of 2"

beyond here there's no way to create a route that's shorter than 3, so it returns none

mev_inspect/arbitrages.py
lukevs: the remaining is also a max though since on the recursion it becomes `max_route_length`
the len of

PR #215, opened 2022-01-05
Switch token amounts for taker and maker on 0x

lukevs:
Token in / out were backwards for 0x swaps

Looking at this example from the tests: https://etherscan.io/tx/0xa043976d736ec8dc930c0556dffd0a86a4bfc80bf98fb7995c791fb4dc488b5d

We see a transfer in of USDC from 0x000.. to the pool (0x7b1) and transfer out of WETH

This switches their orders so they're all right

Docs also say "The address calling this function will be considered the “taker” of the order."
https://protocol.0x.org/en/latest/basics/functions.html#fillrfqorder

Also fixes the issue with this becoming a huge arb:
https://etherscan.io/tx/0xff5390d9d08af7bdfa5d169d72f0cad7f3261ae897e0355d61e100a6e6cff30f


gheise (approved):
👍🏼

PR #217, opened 2022-01-06
Restructure AAVE classifier debt logic

gheise:
Rather than getting the debt data from inputs, this method finds the child transfer of the liquidation call corresponding to the liquidation call and uses this to populate the liquidation. 

This solves some bugs in the db where the inputs of the liquidation call contained maximum values for the debt purchase amount, resulting in enormous losses in the mev_summary materialized views. There is also a correction to two tests, which contained an incorrect debt_purchase_amount for a complicated pair of liquidations: 

- https://etherscan.io/tx/0xf687fedbc4bbc25adb3ef3a35c20c38fb7d35d86d7633d5061d2e3c4f86311b7

and a much simpler one:

- https://etherscan.io/tx/0xedd062c3a728db4b114f2e83cac281d19a9f753e36afa8a35cdbdf1e1dd5d017


mev_inspect/aave_liquidations.py
```diff
@@ -99,3 +100,23 @@ def _get_payback_token_and_amount(
...
+
+
+def _get_debt_data(
+    liquidation: DecodedCallTrace, child_traces: List[ClassifiedTrace], liquidator: str
+) -> Tuple[str, int]:
+    """Get transfer from liquidator to AAVE"""
+
+    for child in child_traces:
+
+        if isinstance(child, CallTrace):
```
lukevs, line 112: DecodedCallTrace is always a call trace

gheise: This check is there for mypy purposes

mev_inspect/aave_liquidations.py
```diff
@@ -99,3 +100,23 @@ def _get_payback_token_and_amount(
...
+
+            child_transfer: Optional[Transfer] = get_transfer(child)
+
+            if child_transfer is not None:
+
+                if child_transfer.from_address == liquidator:
+
+                    return child_transfer.token_address, child_transfer.amount
+
+    return liquidation.inputs["_reserve"], 0
```
lukevs, line 122: Don't we expect this transfer to always exist? Error instead?

mev_inspect/aave_liquidations.py
```diff
@@ -99,3 +100,23 @@ def _get_payback_token_and_amount(
                     return child_transfer.token_address, child_transfer.amount
 
     return liquidation.inputs["_collateral"], 0
```
lukevs, line 102: Same point on erroring here

mev_inspect/aave_liquidations.py
gheise: Wanted to confirm before throwing, since previously had no exceptions for liquidation classifiers

mev_inspect/aave_liquidations.py
lukevs: What's the failure? Because get_transfer takes a CallTrace instead of a DecodedCallTrace?
I think it also needs to be a DecodedCallTrace so better to change that type signature

gheise: ETH transfers are call traces and often show up in the liquidations

lukevs (approved):

PR #219, opened 2022-01-07
Require token amounts in arbitrage swaps to be close to each other

lukevs:
Currently, we check that swaps are joined looking at token addresses and where the tokens are going from / to

In some cases, this leads us to think there's in arb in places where there obviously isn't because the token amounts are wildly different

This PR adds a check that token differences are within 1% of each other
(more than 0 because of potential for leaving dust between swaps)

From looking at large arbs that were incorrect this now filters them out
It also keeps tests passing (except for one reverting test which was also incorrect)


taarushv (approved):

bertmiller:
I'm fine with this change, but we should make a note to investigate if we miss any arbs that right now are wildly unprofitable/profitable and are now simply filtered out.

PR #222, opened 2022-01-08
Include sandwiches that close in arbs

lukevs:
Assume the sandwicher address is where the tokens accumulate

Then allow closing it from either side to be considered the close


gheise (approved):

gheise (commented):
what does closing it from either side mean

PR #223, opened 2022-01-10
Support aave self-liquidations

gheise:
Handles two distinct edge cases:
1. AaveCore contract calls a liquidation which is internally settled and does not result in any transfers:
- There will be debt purchase data from the function inputs but received token amount is 0 since nothing is received

2. Trace contains Reverted/Out of Gas error
- Either debt or received amounts will be 0

This allows the inspector to fall back to the function inputs before throwing

### DIFF
--- a/mev_inspect/aave_liquidations.py
+++ b/mev_inspect/aave_liquidations.py
@@ -14,7 +14,7 @@ from mev_inspect.transfers import get_transfer
 
 def get_aave_liquidations(
     traces: List[ClassifiedTrace],
-) -> Optional[List[Liquidation]]:
+) -> List[Liquidation]:
 
     """Inspect list of classified traces and identify liquidation"""
     liquidations: List[Liquidation] = []
--- a/mev_inspect/aave_liquidations.py
+++ b/mev_inspect/aave_liquidations.py
@@ -14,7 +14,7 @@ from mev_inspect.transfers import get_transfer
 
 def get_aave_liquidations(
     traces: List[ClassifiedTrace],
-) -> List[Liquidation]:
+) -> Optional[List[Liquidation]]:
 
     """Inspect list of classified traces and identify liquidation"""
     liquidations: List[Liquidation] = []
@@ -39,9 +39,15 @@ def get_aave_liquidations(
                 trace, child_traces, liquidator
             )
 
+            if debt_purchase_amount == 0:
+                continue
+
             (received_token_address, received_amount) = _get_received_data(
                 trace, child_traces, liquidator
             )
+
+            if received_amount == 0:
+                continue
 
             liquidations.append(
                 Liquidation(
@@ -63,7 +69,9 @@ def get_aave_liquidations(
 
 
 def _get_received_data(
-    liquidation: DecodedCallTrace, child_traces: List[ClassifiedTrace], liquidator: str
+    liquidation_trace: DecodedCallTrace,
+    child_traces: List[ClassifiedTrace],
+    liquidator: str,
 ) -> Tuple[str, int]:
 
     """Look for and return liquidator payback from liquidation"""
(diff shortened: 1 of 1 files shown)


mev_inspect/aave_liquidations.py
```diff
@@ -63,26 +63,28 @@ def get_aave_liquidations(
 
 
 def _get_received_data(
-    child_traces: List[ClassifiedTrace], liquidator: str
+    liquidation: DecodedCallTrace, child_traces: List[ClassifiedTrace], liquidator: str
```
lukevs, line 66: liquidation implies this is type `Liquidation`

mev_inspect/aave_liquidations.py
```diff
@@ -63,26 +63,28 @@ def get_aave_liquidations(
...
         child_transfer: Optional[Transfer] = get_transfer(child)
 
-        if child_transfer is not None:
-
-            if child_transfer.to_address == liquidator:
+        if child_transfer is not None and child_transfer.to_address == liquidator:
+            return child_transfer.token_address, child_transfer.amount
 
-                return child_transfer.token_address, child_transfer.amount
+    if liquidation.error is not None:
```
lukevs, line 77: this should happen outside this function I think and just return None on the whole liquidation rather than creating dummy data

mev_inspect/aave_liquidations.py
```diff
@@ -63,26 +63,28 @@ def get_aave_liquidations(
...
+        return liquidation.inputs["_collateral"], 0
 
-    raise RuntimeError("Transfer from AAVE to liquidator not found!")
+    else:
+        raise RuntimeError(
+            f"No payback or input data found for liquidation in tx: {liquidation.transaction_hash}"
+        )
 
 
 def _get_debt_data(
```
lukevs, line 86: where does the PR address this part: "AaveCore contract calls a liquidation which is internally settled and does not result in any transfers" - that would still raise an exception after this change right?

gheise: No, this would keep the purchase amount from the inputs and have no received=0

mev_inspect/aave_liquidations.py
gheise: no dummy data is created -  the data is taken from the liquidation call. In the case of Reverted liquidations, this might be useful. If not, we can return None. Why not keep the data given from the inputs for the failed liquidations?

lukevs: i think having the values as 0 is less clear than just not writing the liquidation
I've been leaning that way with swaps and arbs more too

For naive counts of liquidations / swaps etc I think it's better to take the stance of "this liquidation happened" v. "this liquidation half finished"

mev_inspect/aave_liquidations.py
lukevs: in theory they "received" some though right?

for profit calculation should we just then filter out cases where received is 0?

gheise: filter out as in returning None?

mev_inspect/aave_liquidations.py
gheise: Makes sense, I agree

mev_inspect/aave_liquidations.py
lukevs: like when we say "how much MEV $ came from liquidations" what should we do with aave liquidations

mev_inspect/aave_liquidations.py
```diff
@@ -97,10 +101,7 @@ def _get_debt_data(
...
-    if liquidation.error is not None:
-        return liquidation.inputs["_reserve"], liquidation.inputs["_purchaseAmount"]
-
-    else:
-        raise RuntimeError(
-            f"No transfer or input data found for liquidation in tx: {liquidation.transaction_hash}"
-        )
+    return (
+        liquidation_trace.inputs["_reserve"],
+        liquidation_trace.inputs["_purchaseAmount"],
```
lukevs, line 106: can this still be max uint?

mev_inspect/aave_liquidations.py
```diff
@@ -14,7 +14,7 @@
 
 def get_aave_liquidations(
     traces: List[ClassifiedTrace],
-) -> List[Liquidation]:
+) -> Optional[List[Liquidation]]:
```
lukevs: i don't see anywhere where the list itself can be none - should just be empty list if we find none

PR #224, opened 2022-01-10
Support all tokens we have supported for coinbase

lukevs:
cDAI, cUSDC, and DAI are all supported as fetchable from Coinbase, but they weren't in our supported tokens list meaning we won't fetch them

That leads us to not include them in profit calculations


gheise (approved):
Great catch

PR #227, opened 2022-01-11
Added block-list command to enqueue a list of blocks from stdin

RyanRadomski7:
I noticed that my `./mev backfill $start $end` had some missing spots due to errors thrown by my RPC provider. I added this command line tool so I could create a list of missing blocks and enqueue them again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,14 +148,18 @@ For messages permanently failed in the dead letter queue (XQ), query:
 HGETALL dramatiq:default.XQ.msgs
 ```
 
-To query a list of missing blocks in a range, run:
+**Backfilling a list of blocks**
+
+Create a file containing a block per row, for example blocks.txt containing:
 ```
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -148,6 +148,16 @@ For messages permanently failed in the dead letter queue (XQ), query:
 HGETALL dramatiq:default.XQ.msgs
 ```
 
+To query a list of missing blocks in a range, run:
+```
+psql postgresql://postgres:password@localhost:5432/mev_inspect -c "select generate_series(13999471, 13999571) EXCEPT (select distinct block_number from blocks)" --csv -t
```
lukevs, line 153: I think ok to leave it at purely the functionality to backfill a list of blocks since that could come from many sources

ex:
```suggestion

#### Backfilling a list of blocks

Create a file containing a block per row, for example `blocks.txt` containing:
```
12500000
12500001
12500002
```

Then queue the blocks with 
``` 
cat blocks.txt | ./mev block-list
```
```

RyanRadomski7: How about now?

lukevs (approved):

PR #230, opened 2022-01-13
Allow partial results from the db

lukevs:
Currently if the DB fails on fetching any part of data we break out and fetch all

However, this wastes the data we do find which often takes less time than fetching

This PR makes it so we can get partial results from the DB then fetch whatever we don't have

### DIFF
--- a/mev_inspect/block.py
+++ b/mev_inspect/block.py
@@ -1,7 +1,6 @@
 import asyncio
 import logging
-from collections.abc import Awaitable
-from typing import Callable, List, Optional, TypeVar
+from typing import List, Optional, TypeVar
 
 from sqlalchemy import orm
 from web3 import Web3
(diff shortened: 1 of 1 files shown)


mev_inspect/block.py
```diff
@@ -116,7 +113,7 @@ def _find_traces(
     block_number: int,
 ) -> Optional[List[Trace]]:
     result = trace_db_session.execute(
-        "SELECT raw_traces FROM block_traces WHERE block_number = :block_number",
+        "SELECT raw_traces FROM block_traces WHERE block_number = :block_number LIMIT 1",
```
lukevs, line 116: remove once data is clean

gheise (approved):

gheise (approved):

PR #231, opened 2022-01-13
Add coingecko api

gheise:
Simplifies the code for fetching prices and changes the API from Coinbase to CoinGecko to allow for daily price data from all tokens. Limit of 50 calls per minute, no lifetime call limit, more consistent data for each token. Able to achieve max market chart, market cap, and other data with ease


cli.py
```diff
@@ -112,7 +112,7 @@ async def fetch_all_prices():
     inspect_db_session = get_inspect_session()
 
     logger.info("Fetching prices")
-    prices = await fetch_all_supported_prices()
+    prices = fetch_prices()
```
lukevs, line 115: if we're not using async, you can change
```
@cli.command()
@coro
async def fetch_all_prices():
```
to just
```
@cli.command()
def fetch_all_prices():
```

mev_inspect/schemas/prices.py
```diff
@@ -13,11 +16,39 @@
...
+    AAVE_TOKEN_ADDRESS,
+    UNI_TOKEN_ADDRESS,
+    USDC_TOKEN_ADDRESS,
+    DAI_TOKEN_ADDRESS,
+    REN_TOKEN_ADDRESS,
+    CUSDC_TOKEN_ADDRESS,
+    CDAI_TOKEN_ADDRESS,
+]
+
+TOKEN_NAME_BY_ADDRESS = {
```
lukevs, line 32: these names are coingecko specific

since we don't have a `coingecko.py` i think better to name like
```
COINGECKO_NAME_BY_ADDRESS
```

lukevs (approved):

PR #235, opened 2022-01-18
Add LiquidationClassifiers

gheise:


mev_inspect/classifiers/specs/aave.py
```diff
@@ -35,8 +78,36 @@ def get_transfer(trace: DecodedCallTrace) -> Transfer:
...
+    """Look for and return liquidator payback from liquidation"""
+    for transfer in child_transfers:
+
+        if transfer.to_address == liquidator:
+            return transfer.token_address, transfer.amount
+
+    return liquidation_trace.inputs["_collateral"], 0
+
+
+def _get_debt_data(
```
lukevs, line 102: same argument as for compound switching to `Optional[Transfer]`

gheise: good call - will update to match compound

mev_inspect/classifiers/specs/compound.py
```diff
@@ -1,16 +1,76 @@
...
+    def parse_liquidation(
+        liquidation_trace: DecodedCallTrace,
+        child_transfers: List[Transfer],
+        child_traces: List[ClassifiedTrace],
+    ) -> Optional[Liquidation]:
+
+        liquidator = liquidation_trace.from_address
+        liquidated = liquidation_trace.inputs["borrower"]
+        debt_token_address = liquidation_trace.to_address
+        debt_purchase_amount = liquidation_trace.value
```
lukevs, line 27: i'm assuming this is for the cEther case

can you make the assumption more explicit? somehow saying that for cEther we expect value but otherwise we expect inputs / transfer?

mev_inspect/classifiers/specs/compound.py
```diff
@@ -1,16 +1,76 @@
...
+
+        debt_transfer = _get_debt_transfer(liquidator, child_transfers)
+
+        received_transfer = _get_received_transfer(liquidator, child_transfers)
+
+        seize_trace = _get_seize_call(child_traces)
+
+        if debt_transfer is not None:
+            debt_token_address = debt_transfer.token_address
+            debt_purchase_amount = debt_transfer.amount
```
lukevs, line 39: a rational for the above explicitness

if debt_transfer is none then we'll use value even if it's not cEther which could be totally wrong

is there an inputs default to be used here instead?

mev_inspect/classifiers/specs/compound.py
```diff
@@ -1,16 +1,76 @@
...
+        child_transfers: List[Transfer],
+        child_traces: List[ClassifiedTrace],
+    ) -> Optional[Liquidation]:
+
+        liquidator = liquidation_trace.from_address
+        liquidated = liquidation_trace.inputs["borrower"]
+        debt_token_address = liquidation_trace.to_address
+        debt_purchase_amount = liquidation_trace.value
+        received_token_address = liquidation_trace.inputs["cTokenCollateral"]
+        received_amount = 0
```
lukevs, line 29: if 0 is a default meaning "absent" use `None` instead

mev_inspect/classifiers/specs/compound.py
```diff
@@ -1,16 +1,76 @@
...
+        if debt_transfer is not None:
+            debt_token_address = debt_transfer.token_address
+            debt_purchase_amount = debt_transfer.amount
+
+        if received_transfer is not None:
+            received_token_address = received_transfer.token_address
+            received_amount = received_transfer.amount
+
+        elif seize_trace is not None and seize_trace.inputs is not None:
+            received_amount = seize_trace.inputs["seizeTokens"]
```
lukevs, line 46: why not default to the inputs above instead of 0?

mev_inspect/classifiers/specs/compound.py
gheise: yes this for cEther case, will make this more explicit by checking if cEther

mev_inspect/classifiers/specs/compound.py
gheise: yes, repayAmount, will do and separate cToken from cEther

mev_inspect/classifiers/specs/compound.py
gheise: Great call

mev_inspect/classifiers/specs/compound.py
```diff
@@ -1,16 +1,85 @@
...
+        received_token_address = liquidation_trace.inputs["cTokenCollateral"]
+
+        debt_purchase_amount = None
+        received_amount = None
+
+        if debt_token_address == CETH_TOKEN_ADDRESS:
+            debt_purchase_amount = liquidation_trace.value
+
+        else:
+            debt_purchase_amount = liquidation_trace.inputs["repayAmount"]
```
lukevs, line 38: can consolidate to make the type non-optional
```
debt_purchase_amount = (liquidation_trace.value
    if debt_token_address == CETH_TOKEN_ADDRESS
    else liquidation_trace.inputs["repayAmount"]
)
```

lukevs (approved):

lukevs (approved):

PR #238, opened 2022-01-19
Populate prices with listener

lukevs:
Because prices are updated hourly and the mev_summary is updated for each block, there's a chance for prices to be deleted as the summary is being updated causing the data to be incorrect

This removes the existing prices logic and instead makes it a part of the listener so we can control the order of both happening


lukevs:
Changed my mind

PR #239, opened 2022-01-19
Add prices fetch-range

gheise:
## Overview
Add prices-range command that fills prices in a specified time range

## Usage
Fetch prices for all tokens form 01-01-2022 to 01-19-2022:
```
./mev prices fetch-range 1641013200 1642568400
```

## Note
- Timestamps must be in UNIX integer timestamp format
- Minutely data will be used for duration within 1 day, Hourly data will be used for duration between 1 day and 90 days, Daily data will be used for duration above 90 days.

### DIFF
--- a/cli.py
+++ b/cli.py
@@ -119,8 +119,8 @@ def fetch_all_prices():
 
 
 @cli.command()
-@click.argument("after", type=datetime)
-@click.argument("before", type=datetime)
+@click.argument("after", type=click.DateTime(formats=["%Y-%m-%d", "%m-%d-%Y"]))
+@click.argument("before", type=click.DateTime(formats=["%Y-%m-%d", "%m-%d-%Y"]))
 def fetch_range(after: datetime, before: datetime):
     inspect_db_session = get_inspect_session()
 
--- a/cli.py
+++ b/cli.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import sys
+from datetime import datetime
 
 import click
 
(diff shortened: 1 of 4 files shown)


cli.py
```diff
@@ -117,6 +117,19 @@ def fetch_all_prices():
     write_prices(inspect_db_session, prices)
 
 
+@cli.command()
+@click.argument("start", type=int)
+@click.argument("end", type=int)
```
lukevs, line 122: `after` / `before`

`start` and `end` are subjective based on you perspective of time

lukevs: also i think click can deal with type datetime which would be nicer

mev_inspect/prices.py
```diff
@@ -20,7 +20,32 @@ def fetch_prices() -> List[Price]:
...
+                    timestamp=timestamp,
+                    usd_price=token_price,
+                    token_address=token_address,
+                )
+            )
+
+    return prices
+
+
+def fetch_prices_range(start: int, end: int) -> List[Price]:
```
lukevs, line 36: even though this is a "quick" thing, worth the extra 5 mins to make it share logic w/ the one above which is almost identical

cli.py
gheise: coingecko uses UNIX epoch for time, not datetime

mev_inspect/prices.py
gheise: like defining a shared function?

cli.py
lukevs: but people use datetime - not unix timestamp

mev_inspect/prices.py
lukevs: yeah sharing the logic where it's the same

lukevs: or at least the piece to go `prices_data` => `List[Price]`

lukevs (approved):

cli.py
gheise: people, smh

PR #249, opened 2022-01-21
Change nft_trades primary key to include block_number

lukevs:
All main tables should have block number as a primary key

Currently deleting from this table during inspection is our largest CPU usage because it has to search the table each time


gheise (approved):
Tested locally

PR #257, opened 2022-02-04
Fix alchemy endpoint json rpc parameters

IdoBn:
When running the following command (present in the ```README.md```):
```bash
./mev inspect 12914944
```
We get the following output:
```bash
Inspecting block 12914944
Skipping virtualenv creation, as specified in config file.
ERROR:mev_inspect.retry:Request for method eth_getBlockReceipts, params: (12914944,), retrying: 0/5
ERROR:mev_inspect.retry:Request for method eth_feeHistory, params: (1, '0xc51100', None), retrying: 0/5
ERROR:mev_inspect.retry:Request for method eth_feeHistory, params: (1, '0xc51100', None), retrying: 1/5
ERROR:mev_inspect.retry:Request for method eth_getBlockReceipts, params: (12914944,), retrying: 1/5
```
These errors are a result of bad parameters passed to the json-rpc endpoint on some nodes. In my case this was an alchemy node.

### DIFF
--- a/mev_inspect/block.py
+++ b/mev_inspect/block.py
@@ -107,7 +107,7 @@ async def _fetch_block_timestamp(w3, block_number: int) -> int:
 
 
 async def _fetch_block_receipts(w3, block_number: int) -> List[Receipt]:
-    receipts_json = await w3.eth.get_block_receipts(block_number)
+    receipts_json = await w3.eth.get_block_receipts(hex(block_number))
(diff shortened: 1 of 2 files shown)


maxholloway:
In the README it says that the RPC node must support erigon traces and receipts. Does alchemy support erigon receipts? If not, you might try making an RPC endpoint through [pokt.network](https://www.pokt.network).

PR #262, opened 2022-02-11
Add S3 export task 

gheise:
Adds the export to the queue every after inspected block, one block per file for all blocks going forward

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -95,25 +95,25 @@ local_resource(
 )
 
 # if using local S3 exports
-k8s_yaml(secret_from_dict("mev-inspect-export", inputs = {
-    "export-bucket-name" : "local-export",
-    "export-bucket-region": "us-east-1",
-    "export-aws-access-key-id": "foobar",
-    "export-aws-secret-access-key": "foobar",
-}))
+#k8s_yaml(secret_from_dict("mev-inspect-export", inputs = {
+#    "export-bucket-name" : "local-export",
+#    "export-bucket-region": "us-east-1",
+#    "export-aws-access-key-id": "foobar",
+#    "export-aws-secret-access-key": "foobar",
+#}))
 
-helm_remote(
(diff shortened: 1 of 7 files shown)


Tiltfile
```diff
@@ -95,21 +95,25 @@ local_resource(
...
+
+local_resource(
+    'localstack-port-forward',
+    serve_cmd='kubectl port-forward --namespace default svc/localstack 4566:4566',
+    resource_deps=["localstack"]
+)
+
+k8s_yaml(configmap_from_dict("mev-inspect-export", inputs = {
+    "services": "s3",
+}))
```
lukevs, line 119: comment out again

cli.py
```diff
@@ -133,14 +132,6 @@ def fetch_all_prices():
...
 
 
-@cli.command()
-@click.argument("after_block_number", type=int)
-@click.argument("before_block_number", type=int)
-def s3_export(after_block_number: int, before_block_number: int):
-    inspect_db_session = get_inspect_session()
-    export_block_range(inspect_db_session, after_block_number, before_block_number)
-
-
```
lukevs, line 143: we still want a command here. good for one off testing

listener.py
```diff
@@ -82,6 +85,11 @@ async def inspect_next_block(
 
         update_latest_block(inspect_db_session, block_number)
 
+        broker = connect_broker()
+        export_actor = dramatiq.actor(export_block_task, broker=broker)
+        logger.info(f"Sending block {block_number} for export")
```
lukevs, line 90: wrap in a function, but only create the broker / create the actor once

lukevs: otherwise we’re disconnecting and reconnecting to redis a bunch of times when it’s ok to just have the connection

gheise: updated, let me know if that is what you mean

lukevs (approved):

PR #264, opened 2022-02-15
remove cream from compound liquidations classifier

pintail-xyz:
The current implementation of the classifier for crETH liquidations on Cream Finance incorrectly uses Compound's CETH contract address instead of he Cream crETH contract address. This means that blocks containing crETH liquidations cannot be processed by `mev inspect`.

For example, `./mev inspect 13404932` fails with the error:

```
File "/app/mev_inspect/classifiers/specs/compound.py", line 38, in parse_liquidation
    else (liquidation_trace.inputs["repayAmount"], CETH_TOKEN_ADDRESS)
KeyError: 'repayAmount'
command terminated with exit code 1
```

We could introduce Cream-specific logic to the Compound classifier, or create a separate classifier for Cream liquidations, however this is perhaps not worthwhile. According to Etherscan, only 282 liquidation events for crETH have ever occurred. Since [crETH was hacked](https://www.bloomberg.com/news/articles/2021-10-27/defi-protocol-cream-finance-loses-130-million-in-latest-hack) in October 2021 ([exploit transaction](https://etherscan.io/tx/0x0fe2542079644e107cbf13690eb9c2c65963ccb79089ff96bfaf8dced2331c92)), we should not expect any further liquidation events to occur in future. Moreover price data from CoinGecko to determine the value of crETH liquidations is not available.

Therefore this PR simply removes Cream Finance from the Compound classifier.


gheise (commented):
I really like the idea of separating Cream and Compound, but we shouldn't exclude Cream altogether. There's still use to the data from those liquidations besides profit and price metrics. I think for now it would be better to split  them into separate classifiers and have custom logic written for Cream. Overall, great catch and proposal, would love to see a pass at the Cream classifier!

PR #266, opened 2022-02-15
Create low and high priority queues. Put export on high priority, backfill on low

lukevs:
## Issue

We want to export blocks to S3 as we're writing new blocks live

If we're running a large backfill as well, these live exports will be placed on back of the queue

Because the queue is ordered by append time, we have to wait for the backfill to finish before the exports are processed

## The goal

Prioritize live exports over backfills

## This PR

Adds high and low priority queues + priorities

Dramatiq is a little sketchy here.
- Redis backend doesn't have a sense of priority (maybe a good reason to switch to RabbitMQ which does, or even fully to Kafka to skip a step)
- The worker does order by priority for messages it pulls off to work on
- Here we create 2 queue levels - the worker will by default subscribe to messages from both and create separate threads for each
- When it pulls messages off both, it will always prefer messages from the high priority queue

This isn't an incredibly durable solution (if the high priority queue has a ton in it, I don't know that we'll still consistently prioritize it - depends on how often the worker pulls off v. consumes what it has)

But it's sufficient for our use where the high priority is relatively small

## Testing

Using 1 worker locally
Loaded 100 blocks to backfill and let it start progressing
Loaded 3 export blocks to the queue next

Verified that after the current backfill batch was finished, the worker consumed the export tasks then continued backfill

image

### DIFF
--- a/cli.py
+++ b/cli.py
@@ -13,8 +13,10 @@ from mev_inspect.inspector import MEVInspector
 from mev_inspect.prices import fetch_prices, fetch_prices_range
 from mev_inspect.queue.broker import connect_broker
 from mev_inspect.queue.tasks import (
-    BACKFILL_INSPECT_MANY_BLOCKS_PRIORITY,
-    LIVE_EXPORT_BLOCK_PRIORITY,
+    HIGH_PRIORITY,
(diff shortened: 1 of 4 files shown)


lukevs:
@gheise said good

PR #267, opened 2022-02-15
Create pull_request_template.md

lukevs:
Primarily to reduce back and forth when pre-commit is missed

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,16 +1,18 @@
 ## What does this PR do?
 
-A short description of what the PR does and motivation for the PR (what problem does it solve? why is this new feature good?)
+A short description of what the PR does.
 
-## Is this related to an open issue?
+## Related issue
 
-Include the issue number if so
+Link to the issue this PR addresses.
+
+If there isn't already an open issue, create an issue first. This will be our home for discussing the problem itself.
 
 ## Testing
 
 What testing was performed to verify this works? Unit tests are a big plus!
 
 ## Checklist before merging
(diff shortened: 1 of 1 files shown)


pull_request_template.md
```diff
@@ -0,0 +1,15 @@
...
+
+Include the issue number if so
+
+## Testing
+
+What testing was performed to verify this works? Unit tests are a big plus!
+
+## Checklist before merging
+- [ ] I have read the [contributing guide](https://github.com/flashbots/mev-inspect-py/blob/main/CONTRIBUTING.md)
+- [ ] I have installed and run pre-commit hooks
```
gheise, line 15: Add a line that says "All tests pass in the branch"

pull_request_template.md
```diff
@@ -0,0 +1,15 @@
+## What does this PR do?
+
+A short description of what the PR does and motivation for the PR (what problem does it solve? why is this new feature good?)
+
+## Is this related to an open issue?
+
+Include the issue number if so
+
```
gheise, line 8: Should we make this piece mandatory? Aka, if you find a problem, first create the issue, then the PR and tag it, non-optional. In this way the issue will remain open and available while the PR is under review, can still be discused/gather attention

gheise: i rather see the issues first, i think would be useful for us to see what issues are found, keep PR's contextual, and will also help keep the issues up to date

lukevs: I don't feel too strongly about it. If you feel strongly about it let's add it

gheise (approved):

PR #269, opened 2022-02-15
separate out the cream and compound liquidation classifiers

pintail-xyz:
Split out the Cream and Compound liquidation classifiers. Previously the Compound CETH address was erroneously being used by the classifier for cream liquidations such that some blocks containing Cream liquidations would fail (e.g. `./mev inspect 13404932`).


gheise:
Looks good so far, but would like to see some testing that can be cross-checked with Etherscan, with at least a couple different scenarios. Using block 13404932 as a test is a good place to start  (see compound and aave liquidation test)

gheise (changes requested):
Great work! Tests look great. Feel free to request another review if you would like to include additional changes, otherwise I'm happy to merge when you're ready :rocket:

gheise (approved):

PR #270, opened 2022-02-16
Only export empty blocks if there's an existing non-empty one

lukevs:
Currently we write files to S3 for all blocks

This includes blocks with no data, which for small tables is a lot of empty files

This PR changes to only write an empty file if we're communicating that there was data and now there is none


lukevs:
@gheise said good

PR #290, opened 2022-03-22
Support export deletions

gheise:
## What does this PR do?
Add deletions, remove date from filename, and add a created_at field for exports (Requested from Dune for launch)

Tested using localstack and tilt/

## Checklist before merging
- [X] Read the [contributing guide](https://github.com/flashbots/mev-inspect-py/blob/main/CONTRIBUTING.md)
- [X] Installed and ran pre-commit hooks
- [X] All tests pass with `./mev test`

### DIFF
--- a/mev_inspect/s3_export.py
+++ b/mev_inspect/s3_export.py
@@ -38,7 +38,7 @@ def _export_block_by_table(inspect_db_session, block_number: int, table: str) ->
     export_statement = _get_export_statement(table)
     date = round(datetime.utcnow().timestamp())
 
-    object_key = f"{table}/flashbots_{block_number}_{date}.json"
+    object_key = f"{table}/flashbots_{block_number}.json"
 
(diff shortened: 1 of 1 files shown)


elopio (approved):
ok. There are a few things that I don't understand fully, but I won't block this as I don't have the full context.

My suggestions for next time:
- split the pull request in three
- have an issue for each thing that we can use to give further details and discuss. For example, it would be nice to have a trace that clearly indicates that the timestamp was requested by dune, or why are the deletes necessary.
- it's weird that the case we are skipping has the side effect of deleting the key from the bucket. A little comment there could be nice.

For now let's merge, this sounds good and I fully trust your judgement. Also if something goes wrong we can easily solve it later. I would like to continue reviewing pull requests here until I have a better context and until they are so clear that even somebody without context could find it easily.

Thank you Gui!

bhakiyakalimuthu (approved):

PR #292, opened 2022-04-20
Initial JIT Liquidity Classifier Commit

elicbarbieri:
Adding Just in time liquidity classification 

Tests are based on block 13601096
Added necessary classification, schemas and models


mikemcdonald:
@elicbarbieri a couple interesting fields to add to the schema alongside tx details:

- Fee earned by JIT LP
- % of tick range liquidity owned by LP during the sandwich

elicbarbieri:
@mikemcdonald calculating these metrics would require a minimum of 2 calls to the RPC, and probably decoding the logs from the liquidity burn and mint.  All the data from mev-inspect is derived from the transaction traces, and there is no accurate way to get the fees and liquidity across ticks from the traces.  The rest of the codebase minimizes calls to the RPC and it doesn't make sense to change that now.  I will probably pull the raw JIT data into a CSV and calculate these metrics later, and will make sure to post that.

mev_inspect/transfers.py
```diff
@@ -126,3 +126,97 @@ def remove_child_transfers_of_transfers(
         ] = existing_addresses + [transfer.trace_address]
 
     return updated_transfers
+
+
+def get_net_transfers(
+    classified_traces: List[ClassifiedTrace],
+) -> List[Transfer]:
+    """
+    Super Jank...
```
gheise, line 135: Clean up this docstring

mev_inspect/jit_liquidity.py
```diff
@@ -0,0 +1,233 @@
...
+        )
+        if len(bot_trace) == 1 or is_child_trace_address(
+            bot_trace[1].trace_address, bot_trace[0].trace_address
+        ):
+            return _get_bot_address(bot_trace[0], classified_traces)
+        else:
+            return "0x0000000000000000000000000000000000000000"
+
+    # This case is here because from_address is optional in ClassifiedTrace
+    if type(mint_trace.from_address) == str:
```
gheise, line 230: If an attribute (from_address) is Optional in the schema, the object can be created without passing a value and the key will not exist in the dictionary. This will then throw a KeyError. I believe a better check would be 
```if 'from_address' in mint_trace.keys()```

elicbarbieri:
image

PR #295, opened 2022-05-02
Remove quiet flag

gheise:
Internal

Remove quiet flag from listener to make retry verbose

### DIFF
--- a/listener
+++ b/listener
@@ -15,7 +15,6 @@ case "$1" in
         --chdir /app \
         --chuid flashbot \
         --start \
-        --quiet \
         --pidfile $PIDFILE \
         --make-pidfile \
         --startas /bin/bash -- -c "poetry run python listener.py"


Tiltfile
```diff
@@ -15,12 +15,14 @@ helm_remote("redis",
...
+    "secondary_url" : os.environ["SECONDARY_RPC_URL"],
 }))
 
-k8s_yaml(configmap_from_dict("mev-inspect-listener-healthcheck", inputs = {
-    "url" : os.getenv("LISTENER_HEALTHCHECK_URL", default=""),
-}))
+
+#k8s_yaml(configmap_from_dict("mev-inspect-listener-healthcheck", inputs = {
+#    "url" : os.getenv("LISTENER_HEALTHCHECK_URL", default=""),
+#}))
```
elopio, line 26: why is this commented out?

listener.py
```diff
@@ -28,41 +30,35 @@
 # lag to make sure the blocks we see are settled
 BLOCK_NUMBER_LAG = 5
 
+primary_rpc = os.getenv("RPC_URL")
+secondary_rpc = os.getenv("SECONDARY_RPC_URL")
+
+if os.getenv("RPC_URL") is None:
+    raise RuntimeError("Missing primary RPC environment variable: RPC_URL. ")
```
elopio, line 37: this is weird. You do the getenv twice. Maybe better to move it to a function get_rpc_url that is lazy and would fail the first time.
It's also ugly to have these actions run when the module is loaded.

listener.py
```diff
@@ -28,41 +30,35 @@
...
-    healthcheck_url = os.getenv("LISTENER_HEALTHCHECK_URL")
 
     logger.info("Starting...")
 
-    killer = GracefulKiller()
-
-    inspect_db_session = get_inspect_session()
-    trace_db_session = get_trace_session()
+    if _get_inspector_params(rpc_queue[0]) is None and secondary_rpc is not None:
+        rpc_queue.rotate(-1)
```
elopio, line 48: this is not very readable. Sounds too much to have a queue for only two elements. Do you think there will be more possible URLs? If this is only for primary and backup, I would like a clearer solution more specific to take the backup when the primary is not available.

listener.py
```diff
@@ -119,6 +115,39 @@ async def ping_healthcheck_url(url):
...
+        return {
+            "inspector": inspector,
+            "base_provider": base_provider,
+            "killer": killer,
+            "healthcheck_url": healthcheck_url,
+            "inspect_db_session": inspect_db_session,
+            "trace_db_session": trace_db_session,
+            "export_actor": export_actor,
+        }
+    except Exception:
```
elopio, line 147: this is too general. What kinds of exceptions could be thrown here? and why?

elopio:
I know this is work in progress, just wanted to peek a little on how things are going in mev-inspect-py :)

listener.py
gheise: currently investigating

listener.py
elopio: I understand.

PR #298, opened 2022-05-10
Geth additions2

ZigaMr:
## What does this PR do?

Continuation of Supragya Raj's work. This PR adds support for geth nodes by introducing a translation layer between geth and parity traces. Makes use of debug_traceBlockByHash RPC call so make sure to enable debug mode when starting Geth "--http.api debug".

## Related issue

https://github.com/flashbots/mev-inspect-py/pull/112

## Testing

Tested with local node on host machine, used minikube instead of kind. Tested for use with .listener (./mev listener start geth).

## Checklist before merging

- [ ] Read the [contributing guide](https://github.com/flashbots/mev-inspect-py/blob/main/CONTRIBUTING.md)
- [ ] Installed and ran pre-commit hooks
- [ ] All tests pass with `./mev test`

### DIFF
--- a/mev
+++ b/mev
@@ -58,15 +58,17 @@ case "$1" in
 	;;
   inspect)
         block_number=$2
+        rpc_type=$3
         echo "Inspecting block $block_number"
-        kubectl exec -ti deploy/mev-inspect -- poetry run inspect-block $block_number
+        kubectl exec -ti deploy/mev-inspect -- poetry run inspect-block $block_number --type $rpc_type
 	;;
   inspect-many)
-        after_block_number=$2
-        before_block_number=$3
-        echo "Inspecting from block $after_block_number to $before_block_number"
+        start_block_number=$2
+        end_block_number=$3
+        rpc_type=$4
+        echo "Inspecting from block $start_block_number to $end_block_number"
         kubectl exec -ti deploy/mev-inspect -- \
-            poetry run inspect-many-blocks $after_block_number $before_block_number
+            poetry run inspect-many-blocks $start_block_number $end_block_number --type $rpc_type
 	;;
   test)
         shift
--- a/mev
+++ b/mev
@@ -43,7 +43,7 @@ case "$1" in
         redis
     ;;
   listener)
-        kubectl exec -ti deploy/mev-inspect -- ./listener $2
+        kubectl exec -ti deploy/mev-inspect -- ./listener $2 $3
 	;;
   block-list)
         echo "Backfilling blocks from stdin"
--- a/mev_inspect/inspect_block.py
+++ b/mev_inspect/inspect_block.py
@@ -110,7 +110,7 @@ async def inspect_many_blocks(
             w3=w3,
             block_number=block_number,
             trace_db_session=trace_db_session,
-            type=type
+            type=type,
         )
 
         logger.info(f"Block: {block_number} -- Total traces: {len(block.traces)}")
--- a/mev_inspect/inspect_block.py
+++ b/mev_inspect/inspect_block.py
@@ -53,9 +53,7 @@ from mev_inspect.schemas.traces import ClassifiedTrace
 from mev_inspect.schemas.transfers import Transfer
 from mev_inspect.swaps import get_swaps
 from mev_inspect.transfers import get_transfers
-from mev_inspect.liquidations import get_liquidations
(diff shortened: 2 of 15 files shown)


mev_inspect/provider.py
```diff
@@ -1,9 +1,19 @@
 from web3 import AsyncHTTPProvider, Web3
 
+from mev_inspect.geth_poa_middleware import geth_poa_middleware
 from mev_inspect.retry import http_retry_with_backoff_request_middleware
+from mev_inspect.utils import RPCType
 
 
-def get_base_provider(rpc: str, request_timeout: int = 500) -> Web3.AsyncHTTPProvider:
+def get_base_provider(
```
gheise, line 8: Here, rather than checking for the RPC type that's been passed as a variable, it would be beneficial to write a function somewhere that infers the RPC type, maybe from the result of a create_block attempt?

mev
```diff
@@ -43,7 +43,7 @@ case "$1" in
         redis
     ;;
   listener)
-        kubectl exec -ti deploy/mev-inspect -- ./listener $2
+        kubectl exec -ti deploy/mev-inspect -- ./listener $2 $3
```
gheise, line 46: Having to pass the RPC type to the user will become burdensome on every command

mev_inspect/inspect_block.py
```diff
@@ -53,13 +53,15 @@
...
 from mev_inspect.transfers import get_transfers
+from mev_inspect.utils import RPCType
 
 logger = logging.getLogger(__name__)
 
 
 async def inspect_block(
     inspect_db_session: orm.Session,
     w3: Web3,
+    type: RPCType,
```
gheise, line 64: rather than passing the rpc type as an argument, is it possible to retrieve a list of added middlewares from the w3 object itself, and determine the type from the presence/absence of the geth_poa_middleware?

mev_inspect/provider.py
```diff
@@ -1,9 +1,19 @@
...
+    rpc: str, request_timeout: int = 500, type: RPCType = RPCType.parity
+) -> Web3.AsyncHTTPProvider:
     base_provider = AsyncHTTPProvider(rpc, request_kwargs={"timeout": request_timeout})
-    base_provider.middlewares += (http_retry_with_backoff_request_middleware,)
+    if type is RPCType.geth:
+        base_provider.middlewares += (
```
gheise, line 13: what is the advantage of adding middlewares at the provider, rather than using ```Web3.middleware_onion.add``` or ```Web3(middlewares=[geth_middleware])```?

gheise (commented):
Awesome work! Looking forward to your response, some suggestions above.

mev_inspect/provider.py
ZigaMr: What do you propose, to set middlewares at MevInspector constructor? 
I'm not sure why it's set up like that, but for some reason when trying to set it up:
```
base_provider = AsyncHTTPProvider(rpc, request_kwargs={"timeout": request_timeout})
Web3(base_provider, modules={"eth": (AsyncEth,)}, middlewares=[geth_poa_middleware, http_retry_with_backoff_request_middleware,])
```
The function call
`w3.provider.middlewares `
returns empty set, so I'm assuming it has to be done at the provider.

mev_inspect/inspect_block.py
ZigaMr: I agree, will commit new changes

mev
ZigaMr: I agree, will commit new changes

mev_inspect/provider.py
ZigaMr: I agree, will commit new changes

mev_inspect/provider.py
ZigaMr: Hey, yeah it works when enabling debug "--http.api debug" on geth. With a full node I was able to start the listener at few days old blocks (had to update latest_block_update before running listener). 
Listener now infers rpc type from "trace_block" rpc call, but inspect and inspect-many still require the rpc_type parameter.

arsenius-clbs:
So what happened to this PR?

mev_inspect/provider.py
m-r-g-t: I believe there is a cleaner way to do this - namely the web3_clientVersion RPC Call

Issue #283, opened 2022-02-25
Sandwich profit amount inflated in some cases


Dire-0x:
I have found that the `sandwich.profit_amount` in some cases is very inflated... I'll give an example (the amounts below are made up but I have seen many cases of this happening).

frontrun:  1,000,000 USDC -> 100 ETH
backrun: 300 ETH -> 3,200,000 USDC

mev-inspect calculates the profit as 3,200,000 - 1,000,000 USDC

we should calculate the profit pro rata, like this

profit = 100 * (3,200,000 / 300) - 1,000,000

profit should be $66k but is currently calculated as $2.2m

Here is an example case detected by mev-inspect 

frontrun: https://etherscan.io/tx/0x59211262bbc5b8260d4d7ea6311ac81d358a09a99a7422f9a27e3b3fbe92b549
backrun: https://etherscan.io/tx/0x87fda6ee1f2c74e106aab959beca197a8749f988c692aedf10c8499ad5805a81

pmcgoohan:
This is good. I propose a more conservative version which works for imbalances in both directions (overweight front or back runs).

profitFrontrun = (frontOut * (backOut / backIn) - frontIn)
profitBackrun = (backOut - (backIn * (frontIn / frontOut)))
sandwichProfit = min by abs {profitFrontrun, profitBackrun}

tutacrypto:
Hey @Dire-0x @pmcgoohan, what's your progress on this? 

I've had another issue with profit calculation from sandwiches. In this case I get a negative profit (loss), where the transaction looks like it should be positive. Curious to here your thoughts.

Screenshot: 
![Capture d’écran 2022-05-04 à 14 59 32](https://user-images.githubusercontent.com/33679821/166686139-531f42dd-b24f-4e6c-8b92-da36be28e8bd.png)

Frontrun: https://etherscan.io/tx/0x54998a9e51516e5d5979f2a3ea76adb2d609cdeb7c745a01270bf5a03d2cc56a
backrun: https://etherscan.io/tx/0x9d7b0757c6b05751d6141464d4ed7f27499a83445fa17c5e25799675a1cbbeab

You can find the sandwich with this query: 
```
select *
from `flashbots`.`sandwiches`
where block_number = 14659109
```

bryanzk:
@tutacrypto You can check if our calculation was correct here: 
[...]

PR #307, opened 2022-08-04
Adjust sandwiches calculation

gheise:
## What does this PR do?

Adjust the calculation of sandwiches to address incorrect profit amounts. This corrects the naive sandwich profit formula by including an exchange rate to compute only the realized profit for sandwiches with an imbalance between the frontrun and backrun.

## Related issue

- #283 

## Testing

Two test cases included, one for each imbalance. All previously implemented tests for sandwiches pass.

## Checklist before merging
- [x] Read the [contributing guide](https://github.com/flashbots/mev-inspect-py/blob/main/CONTRIBUTING.md)
- [x] Installed and ran pre-commit hooks
- [x] All tests pass with `./mev test`


mev_inspect/sandwiches.py
```diff
@@ -42,32 +42,57 @@ def _get_sandwich_starting_with_swap(
...
             ):
-                sandwiched_swaps.append(other_swap)
+                sandwiched_swaps.append(back_swap)
             elif (
-                other_swap.token_out_address == front_swap.token_in_address
-                and other_swap.token_in_address == front_swap.token_out_address
-                and other_swap.from_address == sandwicher_address
+                back_swap.token_out_address == front_swap.token_in_address
+                and back_swap.token_in_address == front_swap.token_out_address
+                and back_swap.from_address == sandwicher_address
```
come-maiz, line 59: :star: for the rename.

mev_inspect/sandwiches.py
```diff
@@ -42,32 +42,57 @@ def _get_sandwich_starting_with_swap(
...
             ):
                 if len(sandwiched_swaps) > 0:
+                    profit_amount: float
```
come-maiz, line 62: I think this formula needs some documentation in the code. Explain a little what is happening in here, and if you can put a math formula in clearer syntax than python code that would be nice.

@gheise how do we know if the formula is correct?

mev_inspect/sandwiches.py
```diff
@@ -42,32 +42,57 @@ def _get_sandwich_starting_with_swap(
...
+                    if back_swap.token_in_amount == 0 and back_swap.error is None:
+                        raise ValueError("Backrun cannot swap 0 tokens")
```
come-maiz, line 64: This error doesn't have a test case.

gheise: we don't expect this to ever happen on chain; i added this here to be explicit about the assumption that the back swap must be ruled out in the case this value is 0 as to avoid a DivisionByZero error, so test coverage here would not be meaningful

mev_inspect/sandwiches.py
gheise: thanks for pointing that out, the reasons for this are mathematically expressible. I will add further detail on the PR about why i chose this formula and what it means, but i think we should leave this out of the code

Issue #313, opened 2022-10-05
ETH liquidations on Aave not detected


atharvamitragotri:
The Aave spec is not able to detect liquidations for ETH borrowed from the reserve against a collateral token. 

![tx1](https://user-images.githubusercontent.com/49152201/194147053-40715f8d-e49c-4d7a-9018-7c9b9043d075.JPG)

The ETH sent with the liquidation call is not classified as a transfer, hence the whole liquidation is not detected on running ./mev inspect 10384319

![aavescan](https://user-images.githubusercontent.com/49152201/194147098-018a7a95-74cb-4594-8b22-0fe46811fdb0.jpg)

PR #314, opened 2022-10-05
Aave ETH liquidations fixed

atharvamitragotri:
## What does this PR do?

Added a separate transfer classifier spec for the transferToReserve() function on the LendingPoolCore contract. This makes sure that the debt transfer is detected and classified even when the reserve is ETH

## Related issue

#313 ETH liquidations on Aave not detected

## Testing

Unit test for block 10384319 to check if liquidations for ETH are detected. 

## Checklist before merging
- [ ] Read the [contributing guide](https://github.com/flashbots/mev-inspect-py/blob/main/CONTRIBUTING.md)
- [ ] Installed and ran pre-commit hooks
- [ ] All tests pass with `./mev test`


tests/test_aave.py
```diff
@@ -62,6 +63,11 @@ def test_single_liquidation(trace_classifier: TraceClassifier):
     classified_traces = trace_classifier.classify(block.traces)
     result = get_liquidations(classified_traces)
 
+    print("Liquidation result:")
```
sethahrenbach, line 66: Remove print debugs

atharvamitragotri: Removed print debugs

Issue #315, opened 2022-10-05
Reverted liquidiations on Compound not handled properly causing block processing to fail


pintail-xyz:
Reverted Compound liquidations are not properly handled and cause an exception during block processing. For example:

block 15049646 processing fails due to transaction 0x6dd0d8be8a77651f64ef399b47fbc87011bd796b43349c3164ff7da965e0b345

other examples are:

0x4ad1e75751058213ece676f2b68fcccdd8290d5db86539773f8c0b15961e0bc3 (block 15352204)
0x56fe1a490b1cea68957721754c0275b4737ef836ca564c114c051cab88b19938 (block 15382895)
0x779537496f27aa54db49dbcb1c3abc710cf46f9f25a2e4f433db063c6ca5e4ad (block 15382904)
etc

To reproduce run `./mev inspect` with any of the above block numbers.

PR #316, opened 2022-10-09
Fix reverted compound liquidations

pintail-xyz:
## What does this PR do?

It prevents reverted liquidations being sent for parsing (since the classified trace for a reverted transaction will be incomplete and this will cause block processing to fail). Reverted transactions do not result in extraction of MEV and should not be included.

## Related issue

[issue 315](https://github.com/flashbots/mev-inspect-py/issues/315)

## Testing

I've added an additional test to `test_compound.py` this test supplies a block which contains a reverted liquidation transaction and verifies that the transaction does not get included in the set of liquidation traces (and the fact that this block is now processed successfully - it previously didn't - demonstrates the problem has been solved).

## Checklist before merging
- [x] Read the [contributing guide](https://github.com/flashbots/mev-inspect-py/blob/main/CONTRIBUTING.md)
- [x] Installed and ran pre-commit hooks
- [x] All tests pass with `./mev test`


pintail-xyz:
Closing pull request to remove reverted commit from history.

PR #321, opened 2023-01-05
Added uni-v3 and synapse

nikhilsamrat:
Added events for uniswap v3 and synapse protocol, so now arbs can be seen


nikhilsamrat:
opened by mistake peeps sorry

